PR #12, opened 2018-08-20
ThreadContext: ((cur + 1) % 2) == (1 - cur)

markpapadakis:
This should save a few cycles.

### DIFF
--- a/cc/src/core/faster.h
+++ b/cc/src/core/faster.h
@@ -52,15 +52,18 @@ class alignas(Constants::kCacheLineBytes) ThreadContext {
     return contexts_[cur_];
   }
 
+  static_assert((sizeof(contexts_) / sizeof(contexts_[0])) == 2);
+
   inline const ExecutionContext& prev() const {
-    return contexts_[(cur_ + 1) % 2];
+    return contexts_[1 - cur];
   }
+
   inline ExecutionContext& prev() {
-    return contexts_[(cur_ + 1) % 2];
+    return contexts_[1 - cur];
   }
 
   inline void swap() {
-    cur_ = (cur_ + 1) % 2;
+    cur_ = 1 - cur;
   }
 
  private:


markpapadakis:
On Sandy Bridge, Latency of LEA (r,m) is 3 cycles, and AND (r,i) is 1 cycle.

Latency of MOV(r,i) is 1 cycle. Latency of SUB(r,r) is 1 cycle. You save 1 cycle. 
Source: https://www.agner.org/optimize/instruction_tables.pdf

However, those are the minimum/base latency cycles count for those instructions and because the `(1 - cur)` expression is simpler, it should in theory yield even better results depending on the frequency those instructions are evaluated and cache access/invalidation patterns.

I thought it otherwise makes it easier to understand what the code is all about, which is perhaps more important than the potentially saved cycles.

jahunter-m:
Thanks for looking into this. We have not yet seen, in FASTER, single cycles starting to matter, as FASTER spends most (or almost all) of its time stalled on memory accesses, in our experiments.,

I think you have a good point about making the code easier to understand; but I also have a point, as (for now) I have to administer the code base. From my background, I find it easier to think of this swap using modular arithmetic. I assume that, from your background, you find it easier to think of the swap using Boolean arithmetic. People other than us might find it easier to think of the swap as a logical negation; and so on.

The fundamental issue is that C++ does not expose a 0-1 valued datatype, so we must express operations on x in [0, 1] using some higher-level metaphor: like x = (x + 1) % 2, or x = (1 - x), or x = !x. In all of these cases, we intend that x cannot take a value other than 0 or 1; the first and third make this intention explicit, while the second does not. 

In this case, FASTER is swapping two contexts, and it continues to use the old context, as "prev_ctx()". So incrementing (i.e., modular arithmetic) makes sense in this context: logically, the current context is context n, while the previous context is context n-1. Since we only ever need two contexts, it is safe to replace a logical, infinite array of contexts with an array of two. 

For these reasons, I am rejecting this pull request.

markpapadakis:
@jahunter-m : Thank you for looking into it. I appreciate the thoughtful response, and I agree with your reasoning. 
I only looked into FASTER today; I made a quick sweet of the codebase and I really liked what I saw. I plan to spend some more time studying the codebase in order to understand the design and the various implementation specifics. 

One thing that stood out to me was that you are using `io_getevents()` to dequeue a single event from the completion queue, as opposed to multiple events in `QueueIoHandler::TryComplete()`. I think I understand why you do that, but maybe it'd be more efficient to dequeue as many as possible in io_getevents(), store them in some queue/vector and in TryComplete() invoke the callbacks of 1 or more in that queue, and if the queue has been drained, then fill it by invoking io_getevents(). Syscalls are in general not very expensive on Linux nowadays, but they can still be taxing. I will look into it though, because it's likely I am missing something important.

Anyway, thanks for open sourcing FASTER. So many learning opportunities.

jahunter-m:
I appreciate you looking into our code! 

As you have noticed, our io_getevents() loop is not especially efficient.

When thinking about I/O, keep in mind that for random I/O, we are issuing one I/O per record (not per page or batch of records). So FASTER is heavily bound by random I/Os per second. It is probably better to try to complete a batch of I/Os--and probably best to expose this option to the host program, so that its developer can make an intelligent decision.

PR #42, opened 2018-10-05
Changed 'repos' to 'repositories'

Krabbee:


badrishc:
This part of the README is boilerplate text used by many projects, and thus we would rather not modify it.

PR #43, opened 2018-10-08
Updated the official website url in README.md

thareshjose:


badrishc:
Sorry, we are publicizing the short URL and therefore would like to keep the URL as it is.

PR #45, opened 2018-10-08
Fix some problems with garbled / missing API doc comments.

jthelin:
Fix some problems with garbled / missing API doc comments.

Example of compile-time warnings this change fixes:
```
Severity Code Description Project File Line Suppression State

Warning CS1570 XML comment has badly formed XML -- 'Expected an end tag for element 'summary'.' FASTER.core(net46), FASTER.core(netstandard2.0) E:\Depot\GitHub\MSR-FASTER\cs\src\core\Properties\Resources.Designer.cs 65 Active

Warning CS1570 XML comment has badly formed XML -- 'End tag was not expected at this location.' FASTER.core(net46), FASTER.core(netstandard2.0) E:\Depot\GitHub\MSR-FASTER\cs\src\core\Properties\Resources.Designer.cs 83 Active

Warning CS1572 XML comment has a param tag for 'untilAddress', but there is no parameter by that name FASTER.core(net46), FASTER.core(netstandard2.0) E:\Depot\GitHub\MSR-FASTER\cs\src\core\Allocator\PersistentMemoryMalloc.cs 507 Active

Warning CS1573 Parameter 'newSafeReadOnlyAddress' has no matching param tag in the XML comment for 'PersistentMemoryMalloc.OnPagesMarkedReadOnly(long, bool)' (but other parameters do) FASTER.core(net46), FASTER.core(netstandard2.0) E:\Depot\GitHub\MSR-FASTER\cs\src\core\Allocator\PersistentMemoryMalloc.cs 509 Active

```

### DIFF
--- a/cs/src/core/Allocator/PersistentMemoryMalloc.cs
+++ b/cs/src/core/Allocator/PersistentMemoryMalloc.cs
@@ -504,7 +504,7 @@ namespace FASTER.core
         /// Seal: make sure there are no longer any threads writing to the page
         /// Flush: send page to secondary store
         /// </summary>
-        /// <param name="untilAddress"></param>
+        /// <param name="newSafeReadOnlyAddress"></param>
         /// <param name="waitForPendingFlushComplete"></param>
         public void OnPagesMarkedReadOnly(long newSafeReadOnlyAddress, bool waitForPendingFlushComplete = false)
         {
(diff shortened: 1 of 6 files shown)


jthelin:
Good question @badrishc. I played around with things a little more, and I think maybe all the File resources in the .resx can be removed because they are also included as EmbeddedResource's now?
All the test cases seem to run OK when I tried removing everything from the .resx, so do you think that would that be a better solution overall?

https://github.com/Microsoft/FASTER/blob/4b71b84ac09577545770213a233391792ad9700e/cs/src/core/Codegen/CompilerBase.cs#L226

Alternatively, changing the licensing header from line-comments to block-comments provides a temporary workaround too.
```
﻿/* Copyright (c) Microsoft Corporation. All rights reserved.
   Licensed under the MIT license. */
```

PR #46, opened 2018-10-08
Add missing dependency on Microsoft.CodeAnalysis.Analyzers v1.1.0

jthelin:
Add missing dependency on `Microsoft.CodeAnalysis.Analyzers` package, to ensure correct version is restored.

A dependency on package `Microsoft.CodeAnalysis.Analyzers` v1.1.0 implicitly shows up for csharp projects in VS2017, but that package version will not be downloaded / restored on a new machine unless the dependency is called out explicitly in the .csproj file.

### DIFF
--- a/cs/src/core/FASTER.core.csproj
+++ b/cs/src/core/FASTER.core.csproj
@@ -35,9 +35,8 @@
   </PropertyGroup>
 
     <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Scripting">
-      <Version>2.8.2</Version>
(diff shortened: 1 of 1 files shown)


badrishc:
Makes sense, thanks. We also need to modify FASTER.core.nuspec and FASTER.core.debug.nuspec.

jthelin:
> We also need to modify FASTER.core.nuspec and FASTER.core.debug.nuspec.

Won't the Microsoft.CodeAnalysis.* packages only be required at build time, so downstream users of the .nupkg should not require to reference them?

PR #53, opened 2018-10-19
Update LightEpoch.cs

dainmiller:
- Initialize CurrentEpoch with a function for potential use later in file (& more opportunity to document)
- Reformat if statements longer than standard 80 
- Reformat ternary if statement to a standard if statement for readability

### DIFF
--- a/cs/src/core/Epochs/LightEpoch.cs
+++ b/cs/src/core/Epochs/LightEpoch.cs
@@ -104,7 +104,7 @@ namespace FASTER.core
             long p2 = (p + (Constants.kCacheLineBytes - 1)) & ~(Constants.kCacheLineBytes - 1);
             tableAligned = (Entry*)p2;
 
-            CurrentEpoch = 1;
+            CurrentEpoch = _CurrentEpoch()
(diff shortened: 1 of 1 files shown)


badrishc:
The build has failed - see above.

dknippers:
As a random passerby I noticed some things with this PR. I hope it's OK to point these out, if not just ignore this.

- Lots of missing required semicolons, was this PR not built locally at all?
- The method name `_CurrentEpoch` does not seem to follow method naming conventions of the project. 
- `_CurrentEpoch()` does not actually return the current epoch, the name seems incorrect. It should probably be named `_DefaultEpoch` or `_StartingEpoch` or something. If I have an instance of LightEpoch and see this public method, I assume calling it results in the current epoch value, which it does not.
- Why is `_CurrentEpoch()` public?
- Comment of the newly introduced `_CurrentEpoch` function is incomplete.
- `_CurrentEpoch()` is used in the Initialize function, but not in the Uninitialize function where CurrentEpoch is once again set to its default of 1. That would be the actual good use case for it to avoid repeating the "magic" number of 1.

PR #58, opened 2018-10-26
Image Microsoft

VictoriaLa:
Image Microsoft


badrishc:
Thanks for the contribution. The Readme seems fine without the image, so we will consider this in the future.

PR #67, opened 2018-11-25
WIP: Codegen-free FASTER C#

badrishc:
There have been some reports of code gen in FASTER C# causing slow instantiation (5 secs), plus the integration with arbitrary user types has overheads and restrictions (e.g., with access). Therefore, we are experimenting with a code gen free version of FASTER in this PR.

The basic idea is to use calls from [System.Runtime.CompilerServices.Unsafe](https://www.nuget.org/packages/System.Runtime.CompilerServices.Unsafe/) to enable in-place updates of .NET types, and manage the hybrid log without requiring a pointer based core API. I benchmarked Unsafe.AsRef and it appears we can do a billion of these on one core, per second. We are slowly updating parts of the FASTER code, with constant performance benchmarking, to ensure that changes we attempt do not regress performance significantly.

The current status of the PR is: I have remove Key* and Value*, and converted "Functions" to not be static. Performance has not degraded yet. In fact, changing callbacks to user Functions from static to instance-based, has given a 10% hike in performance at scale (72 cores), not sure why.

Comments appreciated - including from @mike-barnett @ReubenBond @gunaprsd.


badrishc:
Work is complete. All tests pass, with no performance regression. Merged, closing PR.

PR #92, opened 2019-02-23
[WIP] Read cache for FASTER

badrishc:
Currently, FASTER supports read caching only by an optional parameter that copies reads to the tail of the hybrid log. However, this results in write amplification in the log, which is undesirable. Further, due to the log growth, it necessitates log compaction even when the workload is read-only.

To address this, we are adding a "read cache" to FASTER. The idea is to have a separately configured circular buffer in memory that will hold recently read items. The FASTER hash index will point to this cache entry, and the entry will in turn point to records in the actual hybrid log.

For simplicity, we will allow at most one cache entry per hash-table entry, and the cache entry will point to the original hybrid log address (tail of record linked-list). This means:
(1) Any upsert or RMW on the same hash entry will result in invalidating the read cache entry and point the hash entry to the newly added hlog entry
(2) A read for a colliding key will either replace the currently cached key or not be cached itself (tbd).

Cache addresses are distinguished from hlog addresses using the most significant bit of the address. When checkpointing the hash table, we will have to replace addresses to cache entries with original addresses.

Adding @peterfreiling  and @gunaprsd for ideas and comments.


cs/src/core/Index/FASTER/FASTERImpl.cs
```diff
@@ -284,22 +304,50 @@ enum LatchOperation : byte
...
+                                       ref readcache.GetValue(newPhysicalAddress));
+            }
+            else
+            {
+                BlockAllocate(recordSize, out newLogicalAddress);
+                newPhysicalAddress = hlog.GetPhysicalAddress(newLogicalAddress);
+                if (UseReadCache && entry.ReadCache)
+                {
+                    entry.Address = readcache.GetInfo(readcache.GetPhysicalAddress(entry.Address & ~Constants.kReadCacheBitMask)).PreviousAddress;
+                }
```
peterfreiling: Dead code, UseReadCache is false here

cs/src/core/Index/FASTER/FASTERImpl.cs
```diff
@@ -1439,6 +1509,33 @@ private void BlockAllocate(int recordSize, out long logicalAddress)
             }
         }
 
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
peterfreiling: >[MethodImpl(MethodImplOptions.AggressiveInlining)] [](start = 8, length = 50)

Can compiler inline a recursive method?

cs/src/core/Index/FASTER/FASTERImpl.cs
```diff
@@ -1439,6 +1509,33 @@ private void BlockAllocate(int recordSize, out long logicalAddress)
...
+                    Thread.Sleep(10);
+                }
+            }
+
+            logicalAddress = logicalAddress < 0 ? -logicalAddress : logicalAddress;
+
+            if (logicalAddress < hlog.ReadOnlyAddress)
+            {
+                Debug.WriteLine("Allocated address is read-only, retrying");
+                BlockAllocateReadCache(recordSize, out logicalAddress);
```
peterfreiling: >BlockAllocateReadCache(recordSize, out logicalAddress); [](start = 16, length = 55)

Does this need to be recursive? What is preventing this from looping infinitely?

cs/test/NativeReadCacheTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            fht = null;
+            log.Close();
+        }
+
+        [Test]
+        public void NativeDiskWriteRead()
+        {
+            InputStruct input = default(InputStruct);
+
+            Random r = new Random(10);
```
peterfreiling: not used?

cs/test/NativeReadCacheTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            for (int i = 0; i < 2000; i++)
+            {
+                var key1 = new KeyStruct { kfield1 = i, kfield2 = i + 1 };
+                var value = new ValueStruct { vfield1 = i, vfield2 = i + 1 };
+                fht.Upsert(ref key1, ref value, Empty.Default, 0);
+            }
+            fht.CompletePending(true);
+
+            r = new Random(10);
+
```
peterfreiling: not used?

cs/test/NativeReadCacheTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            Random r = new Random(10);
+
+            for (int i = 0; i < 2000; i++)
+            {
+                var key1 = new KeyStruct { kfield1 = i, kfield2 = i + 1 };
+                var value = new ValueStruct { vfield1 = i, vfield2 = i + 1 };
+                fht.Upsert(ref key1, ref value, Empty.Default, 0);
+            }
+            fht.CompletePending(true);
```
peterfreiling: Perhaps you should ShiftHeadAddress(LogTailAddress), not only to ensure the synchronicity of the reads is actually due to the read cache but also to reflect the main use case, where the hybrid log in-memory regions will be unnecessary after data load

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -96,10 +98,30 @@ public FasterKV(long size, Functions functions, LogSettings logSettings, Checkpo
...
+                    readcache = new BlittableAllocator<Key, Value>(logSettings, this.comparer);
+                    readcache.Initialize();
+                }
+            }
             else
+            {
                 hlog = new GenericAllocator<Key, Value>(logSettings, serializerSettings, this.comparer);
+                if (UseReadCache)
+                {
+                    readcache = new GenericAllocator<Key, Value>(logSettings, serializerSettings, this.comparer);
```
peterfreiling: >readcache = new GenericAllocator<Key, Value>(logSettings, serializerSettings, this.comparer); [](start = 20, length = 93)

I'm slightly worried about this impact this would have on the per-FASTER-instance footprint. Are there things we can slim down for the read cache allocator, or the original hlog allocator, in this use case?

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -96,10 +98,30 @@ public FasterKV(long size, Functions functions, LogSettings logSettings, Checkpo
...
+                    readcache.Initialize();
+                }
+            }
```
peterfreiling: If the entire data set ends up fitting in the readcache, is there a way to "shrink" the readcache to minimize the required footprint?

cs/src/core/Index/FASTER/FASTER.cs
badrishc: Yes, we should add the ability to provide a "settings" config for the read cache separately from the main hlog allocator (for page size, log size, etc.). This will allow us to size the read cache independently of hlog.

cs/src/core/Index/FASTER/FASTERImpl.cs
```diff
@@ -1439,6 +1509,33 @@ private void BlockAllocate(int recordSize, out long logicalAddress)
             }
         }
 
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void BlockAllocateReadCache(int recordSize, out long logicalAddress)
+        {
+            logicalAddress = readcache.Allocate(recordSize);
```
peterfreiling: >readcache.Allocate(recordSize); [](start = 29, length = 31)

I see the allocate calls on the readcache, but I don't see any cleanup. Where do we manage the circular nature of the readcache?

badrishc: You are right, the cleanup functionality is still not added. This is still WIP (work in progress). We will need the cleanup functionality before we can merge.

cs/src/core/Index/FASTER/FASTERImpl.cs
badrishc: Since this is tail recursion, it should get inlined (I think).

cs/src/core/Index/FASTER/FASTERImpl.cs
peterfreiling: got it

---
In reply to: [259940390](https://github.com/Microsoft/FASTER/pull/92#discussion_r259940390) [](ancestors = 259940390)

cs/src/core/Index/FASTER/FASTERImpl.cs
badrishc: This catches the case where ReadOnlyAddress is updated after allocation but before the allocated address becomes available. Because ReadOnlyAddress is incremented in coarse-grained chunks, it will not recurse forever. Also note that this is tail recursion as before.

cs/test/NativeReadCacheTests.cs
badrishc: Expanded the test to cover this.

cs/src/core/Index/FASTER/FASTER.cs
badrishc: I have made it possible to call ShiftHeadAddress on the read cache. The testcase shows how to do this.
            // Evict the read cache entirely
            fht.ReadCache.ShiftHeadAddress(fht.ReadCache.GetTailAddress());

cs/src/core/Index/FASTER/FASTER.cs
badrishc: Added settings config for read cache, and also allow users to control the read cache separately.

cs/src/core/Index/FASTER/FASTERImpl.cs
badrishc: Cleanup functionality (beta) added to the PR.

badrishc:
New iteration of code pushed.

cs/src/core/Allocator/GenericAllocator.cs
```diff
@@ -251,14 +266,12 @@ protected override void ClearPage(long page)
                         IOCompletionCallback callback, PageAsyncFlushResult<TContext> asyncResult,
                         IDevice device, IDevice objlogDevice, long intendedDestinationPage = -1, long[] localSegmentOffsets = null)
         {
-            /*
-            if (!(KeyHasObjects() || ValueHasObjects()))
+            // Short circuit if we are using a null device
```
peterfreiling: >// Short circuit if we are using a null device [](start = 12, length = 46)

Nice, was going to suggest this after investigating the previous bugs

cs/src/core/Index/FASTER/LogAccessor.cs
```diff
@@ -0,0 +1,125 @@
...
+        public LogAccessor(IFasterKV<Key, Value, Input, Output, Context> fht, AllocatorBase<Key, Value> allocator)
+        {
+            this.fht = fht;
+            this.allocator = allocator;
+        }
+
+        /// <summary>
+        /// Tail address of log
+        /// </summary>
+        public long TailAddress => allocator.GetTailAddress();
```
peterfreiling: We will need to expose the other addresses as well.

PR #96, opened 2019-02-25
Alternative hash entry cleanup, best-effort delete of hash entries in memory

badrishc:
When begin address moves forward, instead of scanning entire hash table to remove outdated entries, we piggyback the cleanup on insert operations.

We also added preliminary support for best effort delete (DeleteFromMemory). The functionality is used to clean up hash entries if possible (when deleted record in in memory and hash entry points directly to it). This is useful to clean up hash table entries in case records are in memory but are deleted, therefore allowing us to reuse the corresponding hash entries. Whether this should be exposed as a public API is still under consideration.


badrishc:
Closing as this PR has been merged into the readcache PR at https://github.com/Microsoft/FASTER/pull/92

PR #103, opened 2019-02-28
Changes making "Faster" file storage support unix folder paths

vibeeshan025:
All changes are related to Directory Seperator charector, Windows slash is converted to unix compatible code using `Path.DirectorySeparatorChar` which is platform independant.

I would like to contribute in .Net Core support, Mac OS and Linux areas. Like converting playground area to unix compatible code etc.


badrishc:
Thanks for your contribution, and look forward to more in future as well.

Issue #70, opened 2018-12-15
Compaction based garbage collection


badrishc:
Currently we support garbage collection based on expiry of keys, via truncation from the head of the log. This has extremely low overhead, and captures the cases where there is a limit on TTL of all keys. However, there are cases where actual compaction of live keys is necessary. Thus, we need an mechanism to roll forward pages from the head, by determining liveness of keys in the pages to be expired, and reinserting live keys at the tail. Adding @gunaprsd.

gunaprsd:
It is impossible to avoid scanning through the records (to check the key) when you use an approximate summary of keys such as bloom filter. One alternative is to store the keys in a separate file alongside the hybrid log. This can also be computed as an offline post-processing. 

For every page or segment we have a keys file. We wish to garbage collect `segments[0]` and the subsequent segments are `segments[1], ..., segments[n]'. 

## Hash-Table Based Solution
The only requirement on the keys file is that it must contain all keys that have a record in the corresponding hybrid log segment. 

```Csharp
hashTable = create hash table from segments[0].keys file to {valid, invalid}
for(int i = 1; i <= n; i++) {
  other = segments[i].keys;
  foreach(var key in other) {
    if(hashTable[key] == 'valid') {
      hashTable[key] = 'invalid';
    }
  }
}

foreach(var key in hashTable.keys) {
  if(hashTable[key] == 'invalid') {
    create corresponding record at tail;
  }
}
```

## Sort-Merge-Diff Solution
Another variant is to have a sorted keys file for each segment. For each `segments[i]` with `i > 0`, we scan through sorted keys and merge-diff with that of `segments[0]`

```Csharp
res = segments[0].sorted_keys;
for(int i = 1; i <= n; i++) {
 seg = res;
 res.clear();
 other = segments[1].sorted_keys;
 seg.begin();
 other.begin();
 while(seg.hasNext() && other.hasNext()) {
   if(seg.front() == other.front()) {
    seg.next();
    other.next();
   } else if (seg.front() < other.front()) {
    res.add(seg.front());
    seg.next();
   } else {
    other.next();
   }
 }
 while(seg.hasNext()) {
  res.add(seg.next());
 }
 seg.close();
 other.close();
 res.close();
}
```

## Comparison
|                | Hash-Table-Based                         | Sort-Merge-Diff                                                    |
|----------------|------------------------------------------|--------------------------------------------------------------------|
[...]

PR #112, opened 2019-03-20
[WIP] FASTER Log Compaction

badrishc:
Until now, the only way to clean up the log has been via log truncation (using `ShiftBeginAddress`), which has the model that older keys expire when they fall off the head of the log (if they have not been updated after that point).

Log compaction introduces the ability to scan the head of the log, until the specified address, for the keys that will get deleted in that range. Then, it scans the rest of the log to check which keys are live. Live keys are then copied over to the tail of the log, resulting in compaction by eliminating deleted or subsequently updated keys.

Log compaction will be exposed via an API call on FASTER as follows:
```cs
   fht.Log.Compact(compactUntil)
```

Our first iteration of the feature is single threaded, so that log compaction can be run as a background process. We will not support parallel instantiation of multiple log compaction threads. Log compaction is blocking, i.e., it returns after compaction until the specified address is complete, and `BeginAddress` has been shifted to `compactUntil`.

Also refer to issue https://github.com/Microsoft/FASTER/issues/70.

Not yet ready for merge or review, but adding @gunaprsd and @peterfreiling for comments.

### DIFF
--- a/cs/test/BlittableLogCompactionTests.cs
+++ b/cs/test/BlittableLogCompactionTests.cs
@@ -23,7 +23,7 @@ namespace FASTER.test
         [SetUp]
         public void Setup()
         {
-            log = Devices.CreateLogDevice(TestContext.CurrentContext.TestDirectory + "\\hlog4.log", deleteOnClose: true);
+            log = Devices.CreateLogDevice(TestContext.CurrentContext.TestDirectory + "\\BlittableLogCompactionTests.log", deleteOnClose: true);
             fht = new FasterKV<KeyStruct, ValueStruct, InputStruct, OutputStruct, int, FunctionsCompaction>
                 (1L << 20, new FunctionsCompaction(), new LogSettings { LogDevice = log, MemorySizeBits = 15, PageSizeBits = 7 });
             fht.StartSession();
--- a/cs/test/BlittableLogCompactionTests.cs
+++ b/cs/test/BlittableLogCompactionTests.cs
@@ -17,15 +17,15 @@ namespace FASTER.test
     [TestFixture]
     internal class BlittableLogCompactionTests
     {
-        private FasterKV<KeyStruct, ValueStruct, InputStruct, OutputStruct, Empty, Functions> fht;
+        private FasterKV<KeyStruct, ValueStruct, InputStruct, OutputStruct, int, FunctionsCompaction> fht;
         private IDevice log;
 
         [SetUp]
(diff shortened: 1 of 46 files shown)


gunaprsd:
I have not talked about concurrency between updates and reinsertion in https://github.com/Microsoft/FASTER/issues/70

The reinsertion algorithm needs to check for updates post the time we checked a record  for GC. A straightforward way is to record tail address at start and always check the record chain after this address before doing the CAS for reinsertion.

badrishc:
Here is my plan for handling the final case (NYI). We first complete a scan until SafeReadOnly (may need to continue the scan in case SRO has moved forward in the interim during the main log scan).

Then, for each live key to be re-inserted, I will perform a range-limited TraceBack until that SRO point, in order to check if the key exists in the region between tail and that point. If not found, we will perform a CAS to install the key.

I was thinking we cannot use "tail address" above instead of SRO, because the log may just be getting allocated and the record may be in an unstable state beyond SRO. In general, it is unsafe to scan beyond SRO because records in that region of memory may not yet have been successfully upserted, i.e., chained into the hash table.

The Log Scan capability has made compaction significantly simpler, and I am using an in-memory instance of FasterKv for temporarily storing keys to apply the validity test.

cs/test/BlittableLogCompactionTests.cs
```diff
@@ -0,0 +1,136 @@
...
+                var value = new ValueStruct { vfield1 = i, vfield2 = i + 1 };
+
+                var status = fht.Read(ref key1, ref input, ref output, Empty.Default, 0);
+                if (status == Status.PENDING)
+                    fht.CompletePending(true);
+                else
+                {
+                    Assert.IsTrue(status == Status.OK);
+                    Assert.IsTrue(output.value.vfield1 == value.vfield1);
+                    Assert.IsTrue(output.value.vfield2 == value.vfield2);
```
peterfreiling: Shouldn't these be outside of the "else" ? I.e., validate output for async Reads as well?

badrishc: Async read results are checked in the ReadCompletionCallback in FunctionsCompaction.

badrishc:
We now have a fully working prototype that supports deletes and log compaction. We will keep testing and cleaning up to bring the support up to production quality before merging. The API for log compaction also needs some thinking, e.g., do we provide a fixed memory budget or let users specify a log head range to compact from.

PR #117, opened 2019-04-03
Removing statics from codebase

badrishc:
We have used ThreadStatic until now, for storing thread-local variables. This does not work when you have multiple instances of FASTER operating at the same time. This PR fixes the problem by using instance-specific thread-local variables instead. We have implemented our own version of thread-local for this purpose, called FastThreadLocal<T>. Incidentally, FastThreadLocal<T> is optimized to be around 10x faster than ThreadLocal<T> in our micro-benchmarks. There is a configurable (in code) limit of 128 instances of a particular thread-local variable type. This may be relaxed in future if the requirement comes up. With this PR, there are no shared objects between instances, except for the buffer pool which is shared static for the entire process.

### DIFF
--- a/cs/src/core/Epochs/LightEpoch.cs
+++ b/cs/src/core/Epochs/LightEpoch.cs
@@ -52,7 +52,7 @@ namespace FASTER.core
         /// <summary>
         /// A thread's entry in the epoch table.
         /// </summary>
-        private ThreadLocal<int> threadEntryIndex;
+        private FastThreadLocal<int> threadEntryIndex;
 
         /// <summary>
         /// Global current epoch value
@@ -74,20 +74,12 @@ namespace FASTER.core
         }
 
         /// <summary>
-        /// 
-        /// </summary>
-        ~LightEpoch()
-        {
-            Uninitialize();
-        }
-
-        /// <summary>
         /// Initialize the epoch table
         /// </summary>
         /// <param name="size"></param>
         unsafe void Initialize(int size)
         {
-            threadEntryIndex = new ThreadLocal<int>();
+            threadEntryIndex = new FastThreadLocal<int>();
             numEntries = size;
 
             // Over-allocate to do cache-line alignment
@@ -110,7 +102,7 @@ namespace FASTER.core
         /// <summary>
         /// Clean up epoch table
         /// </summary>
-        void Uninitialize()
+        public void Dispose()
         {
             tableHandle.Free();
             tableAligned = null;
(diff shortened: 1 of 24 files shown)


badrishc:
> `FasterThreadLocal<T>.InitializeThread` isn't thread-safe. It is being called in `LightEpoch.Acquire` and can lead to concurrency issues.

Fixed as follows:

```cs
            if (values == null)
            {
                Interlocked.CompareExchange(ref values, new T[kMaxInstances], null);
            }
```

> Can you explain how the current implementation of `FastThreadLocal<T>` is instance specific? I am confused because the values array is thread static which I suppose means that two objects of `FastThreadLocal<int>` access the same values array if done from the same thread.

FastThreadLocal has a thread-static array, i.e., each thread sees its own copy of the array. Each _instance_ of the object (regardless of threads) will correspond to a unique entry offset (`id`) in the array. You can visualize this as a 2-d matrix of threads and instances, with one array per thread.

peterfreiling:
> > `FasterThreadLocal<T>.InitializeThread` isn't thread-safe. It is being called in `LightEpoch.Acquire` and can lead to concurrency issues.
> 
> Fixed as follows:
> 
> ```cs
>             if (values == null)
>             {
>                 Interlocked.CompareExchange(ref values, new T[kMaxInstances], null);
>             }
> ```
Why is this necessary for a thread static variable? Won't each thread sees a different copy of "values", making the Interlocked operation unnecessary?

badrishc:
> > > `FasterThreadLocal<T>.InitializeThread` isn't thread-safe. It is being called in `LightEpoch.Acquire` and can lead to concurrency issues.
> > 
> > 
> > Fixed as follows:
> > ```cs
> >             if (values == null)
> >             {
> >                 Interlocked.CompareExchange(ref values, new T[kMaxInstances], null);
> >             }
> > ```
> 
> Why is this necessary for a thread static variable? Won't each thread sees a different copy of "values", making the Interlocked operation unnecessary?

Correct, it's not needed. My original code is fine, I think.

cs/src/core/Epochs/LightEpoch.cs
```diff
@@ -132,23 +121,17 @@ void Uninitialize()
         /// <returns>Result of the check</returns>
         public bool IsProtected()
         {
-            return (kInvalidIndex != threadEntryIndex);
+            return kInvalidIndex != threadEntryIndex.Value;
```
peterfreiling: >threadEntryIndex.Value; [](start = 36, length = 23)

This can be called from an async callback thread that has not initialized itself, so we should check to see if it's initialized before derefing the value

PR #120, opened 2019-04-18
Variable length struct support in FASTER C#

badrishc:
The title appears contradictory, but bear with us for a minute.

We are (re-)introducing an unsafe advanced use case for FASTER, where there is a need to
store and index variable length keys and/or values in the hybrid log, without using classes
or a separate object log.

The basic idea is to use `ref struct` as a proxy for pointers to variable-sized memory in 
C#. These objects are placed contiguously in-place in the single hybrid log, leading to 
efficient packing while avoiding the additional I/O (on reads and writes) that a separate 
object log entails. Users have to be careful that updates do not attempt to increase the 
size of the object, if the object is already present in the mutable region.

Users provide information on the actual length of the data underlying the types, by 
providing implementations for an IVariableLengthStruct<T> interface. Serializers are not 
required, as these are effectively value-types.

This interface to FASTER is unsafe -- one may orthogonally create safe APIs on top of this 
raw functionality using, for example, `Span<T>` and `Memory<T>`.

See an example of how this functionality is used here: https://github.com/Microsoft/FASTER/tree/varlenstructs/cs/playground/VarLenStructSample


watkinsm:
@badrishc: I hope you found the impact report helpful! I came across FASTER and requested the report just out of curiosity, but if you're interested we can set up @softagram-bot to analyze all of your open source pull requests for free. On your end it would just require setting up a webhook, and we'll take care of the rest. Let me know 🙂

PR #123, opened 2019-04-24
cc: benchmark: map arg to workload enum

matthewbrookes:
What:
* use a map between possible workload strings and enum values

Why:
* currently, regardless of the supplied argument it runs the YCSB_A_50_50
benchmark. By mapping possible arguments to the desired workload, can
ensure the correct workload is run.

### DIFF
--- a/cc/benchmark-dir/benchmark.cc
+++ b/cc/benchmark-dir/benchmark.cc
@@ -8,6 +8,7 @@
 #include <cstdlib>
 #include <random>
 #include <string>
+#include <map>
 
 #include "file.h"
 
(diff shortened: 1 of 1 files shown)


badrishc:
The argument supplied to the program is an integer, like 0, 1, etc. that will get mapped to the enum (I think). We would like to keep the benchmark simple and avoid introducing a data structure such as map here, if we can avoid it. Perhaps this can be resolved by better documentation of the args.

matthewbrookes:
Hey,
Yeah I can see that's the intention. I'll close this PR and open one with some improved documentation because I was a bit confused over this.
Anyways, thanks for taking a look :)

PR #130, opened 2019-05-21
Support async/await session model

### LINKED ISSUE #140: Better support for async/await threading model
The FASTER model is based on a set of threads, each of which can start a session with FASTER, perform a sequence of operations, and later stop the session. Operations consist of bursts of activity, with periodic invocations of a Refresh() call to FASTER in between. It would not be okay for a thread/session to go to sleep (i.e., stop calling Refresh) without stopping its session. Sessions and threads are tightly integrated in this model.

However, this tight coupling of threads and sessions in FASTER is not ideal for an async/await API using the C# thread pool. This PR aims to investigate and prototype alternatives to this approach, and implement the best one. There is next to no code right now, this PR for now serves as a place to discuss options and alternatives.

See https://github.com/microsoft/FASTER/pull/130 for some more discussion and a very early work-in-progress PR.

badrishc:
The FASTER model is based on a set of threads, each of which can start a session with FASTER, perform a sequence of operations, and later stop the session. Operations consist of bursts of activity, with periodic invocations of a Refresh() call to FASTER in between. It would not be okay for a thread/session to go to sleep (i.e., stop calling Refresh) without stopping its session. Sessions and threads are tightly integrated in this model.

However, this tight coupling of threads and sessions in FASTER is not ideal for an async/await API using the C# thread pool. This PR aims to investigate and prototype alternatives to this approach, and implement the best one. There is next to no code right now, this PR for now serves as a place to discuss options and alternatives.

Adding @JorgeCandeias @gunaprsd @ReubenBond for comments.

Fix #140


badrishc:
The first option is to eliminate thread-local variables in FASTER entirely, and have logical sessions that may move across threads. This will allow FASTER to provide persistence guarantees for a logical session (e.g., commit/persist all operations with LSN < k on the _logical_ session), regardless of which thread the session is currently executing on. This means users can seamlessly use other async operations in their code.

Every FASTER operation would then need a session ID parameter, which is a bit ugly. Perhaps the default could still be “threads==sessions” behavior by using a session-free overload, where a thread-local variable is used to store the default session ID.

Currently, FASTER operations that go async are affinitized back to the thread that issued the I/O. The client is supposed to call CompletePending to "resume" these operations on that thread. We could leave this as it is, with logical sessions expected to call CompletePending to complete the I/Os on that session. However, it may instead be interesting to support the async model for these operations, where the operation returns a Task instead.

Internally, we could use `TaskCompletionSource` to handle the async operations. We could create a simplified async API layer for FASTER that simply uses FASTER’s `Context` generic parameter to store the TaskCompletionSource. An async callback can then complete the task using this context. Since we are in async world, there is no need for a separate user-facing `Context` anyway.

Note that switching a session from one thread to another is generally expensive, as the session context needs to move to the other thread. However, ideally, this would happen rarely. In the common synchronous case, we would expect a thread to have a single session to FASTER, leading to high performance very similar to the current thread-local session model.

itn3000:
How would you like to use `ValueTask` + `IValueTaskSource`?
It is used in library which is required high performance like `System.Threading.Channels` and `System.IO.Pipelines`.
It may be more efficient than using `TaskCompletionSource`.

JorgeCandeias:
> The first option is to eliminate thread-local variables in FASTER entirely, and have logical sessions that may move across threads. This will allow FASTER to provide persistence guarantees for a logical session (e.g., commit/persist all operations with LSN < k on the _logical_ session), regardless of which thread the session is currently executing on. This means users can seamlessly use other async operations in their code.
> 
> Every FASTER operation would then need a session ID parameter, which is a bit ugly. Perhaps the default could still be “threads==sessions” behavior by using a session-free overload, where a thread-local variable is used to store the default session ID.

One option to do away with threads could be for the lookup to pool sessions, renting them out and taking them back as needed. That could enable user syntax such as this:

``` csharp
using (ISession session = lookup.OpenSession())
{
    await session.UpsertAsync( /* stuff here */ )
}
```

In the model above, the `ISession` is just a shell for an object that contains a unique session id and a reference to the owning lookup. Calls on the session translate to calls on the lookup using the actual session. Disposing `ISession` returns the underlying rented session back to the owning session pool, while ensuring `Refresh()` happens at some point.

> 
> Currently, FASTER operations that go async are affinitized back to the thread that issued the I/O. The client is supposed to call CompletePending to "resume" these operations on that thread. We could leave this as it is, with logical sessions expected to call CompletePending to complete the I/Os on that session. However, it may instead be interesting to support the async model for these operations, where the operation returns a Task instead.
> 
> Internally, we could use `TaskCompletionSource` to handle the async operations. We could create a simplified async API layer for FASTER that simply uses FASTER’s `Context` generic parameter to store the TaskCompletionSource. An async callback can then complete the task using this context. Since we are in async world, there is no need for a separate user-facing `Context` anyway.
> 

This what the [Orleans sample](https://github.com/dotnet/orleans/pull/5618) is doing now. The drawback is that we now need an extra allocation for every operation, even those completing on the sync path. At this time is hard to tell what performance difference it makes due to the overhead from everything else. An implementation using `ValueTask` and a pooled task factory may have some merit.

> Note that switching a session from one thread to another is generally expensive, as the session context needs to move to the other thread. However, ideally, this would happen rarely. In the common synchronous case, we would expect a thread to have a single session to FASTER, leading to high performance very similar to the current thread-local session model.

I think the pooled session model above could work, by making the session pool return the session that is affinitized to the running thread (instead of just an id) and only create a new one if it does not exist yet. So it would be more of a `GetOrCreate()`-style lookup than an actual pool. I suspect there would be some overhead as the ThreadPool adds and removes threads but in the case of Orleans that would be amortized down as the system lives on.

I'll take a stab at the `AsyncFasterKV` in this branch to see if this could work.

JorgeCandeias:
> How would you like to use `ValueTask` + `IValueTaskSource`?
> It is used in library which is required high performance like `System.Threading.Channels` and `System.IO.Pipelines`.
> It may be more efficient than using `TaskCompletionSource`.

Indeed, this may have some merit, as operations in FASTER can complete sync or async.

badrishc:
* The idea of an ISession is nice. User-facing sessions are Guids, and may need to be mapped to internal "sessions" which will simply be offsets in the epoch table. A session abstraction can potentially avoid an explicit map between session Guid and the table offset.

* Yeah, the extra allocation is not good. We really need the normal sync code path to be as lightweight as it is right now. One crazy idea is to make context an out param, allowing a user defined callback to create the context when it is needed.

> I think the pooled session model above could work, by making the session pool return the session that is affinitized to the running thread (instead of just an id) and only create a new one if it does not exist yet. 

* If the pooled session was simply going to return the session associated with the thread, then what would happen after an async call returns on a different thread? Would the code continue to use that same old session, now on a different continuation thread? What about another session that now tries to continue on the older thread? It seems that we would need sessions to be independent of threads, right?

JorgeCandeias:
> One crazy idea is to make context an out param, allowing a user defined callback to create the context when it is needed.

At least this can get the shell `AsyncFasterKV` working early without affecting the sync path or the core `FasterKV` too much.

> If the pooled session was simply going to return the session associated with the thread, then what would happen after an async call returns on a different thread? Would the code continue to use that same old session, now on a different continuation thread? What about another session that now tries to continue on the older thread? It seems that we would need sessions to be independent of threads, right?

You're right, that would open a can of worms, if it would even work at all. Best to split those concerns from the onset.

As a first step, I'd just exploit the context parameter as the Orleans sample does it, and either make the context on the `FasterKV` an `out` or a `ref` to allow delaying the allocation up to when we hit the async path. That will be a breaking change on the current surface, but at least we can test it. And while a `ValueTask` sync path may have merit, the problem is that we can only await on it once. That's fine for tight-controlled internals, but prone to bugs on a public surface. Let's see how far we can get with plain old `Task`s.

ReubenBond:
I think the pooled session model is appropriate. The sessions would essentially be async-local instead of thread-local.

Alternatively, each grain could rent a session at activation and return it at deactivation. That would require something to pulse `Refresh()`, which we could potentially add hooks for (or use a timer).

Thread-local has issues with a dynamically sized thread pool such as .NET's ThreadPool (which grains do not use currently)

badrishc:
Okay, I'm just getting back to this topic after some travel. I plan to work on decoupling threads and sessions in Faster now, on this branch.

The epoch table will contain one row per session. There will be no thread local variables any more. When someone closes a session, the epoch table entry will become available for another session to rent out.

@reubenbond, the session is a local variable for the application and thus effectively 'async local'. Is this what you meant, or were you thinking of [AsyncLocal<T>](https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=netframework-4.8)? The latter seems not very useful to me, perhaps I am missing something ...

badrishc:
Rough sketch of session checked in, without touching the existing thread-local framework: session-based access will just update existing thread-local variables using the session's values, before proceeding with usual operation normally.

badrishc:
@JorgeCandeias -- I added some preliminary support for pooled sessions. we need to see if this can improve the Orleans integration. See samples at https://github.com/microsoft/FASTER/blob/async-support/cs/test/SessionFASTERTests.cs

* App can request a session and perform FASTER operations on it as it hops across threads (e.g., due to other async calls it might make)
* Two apps can request two different sessions on the same thread.
* An app can dispose a session, i.e., the next app that requests a session will not resume that session.
* An app can "return" a session to FASTER. In this case, another app can call GetSharedSession and potentially get back the same session to continue operations on.
* Apps can always call CreateSharedSession to get a fresh session

I'm still a bit fuzzy on how exactly this would map to the Orleans threading model, and would appreciate your thoughts. E.g., since individual grains may perform other async ops at any time, they may be holding an active FASTER session when they go async, right? In that case, when a different grain is scheduled on that thread, it would need to "rent" a different session. Thus, the number of active sessions may grow to more than the number of threads -- it would depend on the number of pending activated grains.

badrishc:
@JorgeCandeias -- any comments? Would this version be of interest to you to improve the FASTER-Orleans sample? And what do you think of adding a true async interface to FASTER?

JorgeCandeias:
Hi @badrishc, many apologies for the delay, other things moved up on my list in the meantime, but this one is still on it.

Yes, this version is already useful for improving the sample, so I'll take it on-board, thank you!

That said, being a synchronous interface, it will still have to run on the thread pool in order not to limit throughput on the Orleans task scheduler when spinning for I/O completions. A true async interface with no blocking or background threads will enable direct integration with the Orleans scheduler and therefore not having to rely on the thread pool.

More importantly, it will become easier to promote this to product developers looking for straightforward solutions. 😄 

---

> E.g., since individual grains may perform other async ops at any time, they may be holding an active FASTER session when they go async, right? In that case, when a different grain is scheduled on that thread, it would need to "rent" a different session. Thus, the number of active sessions may grow to more than the number of threads -- it would depend on the number of pending activated grains.

This is correct. It has less to do with grains and more with TPL tasks. Although the Orleans scheduler only creates as many worker threads as processors, there can be great many tasks pending in the system, due to awaiting something non-CPU bound to complete, like I/O operations. In the meantime other tasks will run.

For background on this, Orleans does this to keep the CPU *doing actual user work* as much as possible and avoid wasting cycles on context switching or spinning for I/O. The reason is that grain *requests* on Orleans tend to be very fine-grained on the CPU, e.g. in the microseconds and lower (even if they *await* I/O tasks for longer), so having an Orleans worker thread spinning even for a few milliseconds can lower the throughput ceiling. Orleans is opinionated on this and enforces responsiveness in the system by killing any task taking over 30s and issuing warnings when tasks take over 200ms, regardless of CPU usage.

In my head, for an async/await model to work with the Orleans scheduler, I think...

* For isolated sessions, there must be no (or very high) upper bound for the number of in-use sessions against a lookup, even if the pool itself has some limit. If faster can support this, then the sessions do not need to be thread-safe. Each grain task can request one, do work, go async, do more work, then return the session to the session pool for another task to request. This assumes developers won't pass around sessions between threads, just like we don't pass `SqlConnection`s around.

* For shared sessions, the sessions will need to be thread-safe. Let's say task 1 requests session A on thread 1 and awaits, then task 2 requests session A again on thread 1 and also awaits. When either continuation resumes, there's no guarantee the work will be performed on the same thread, as far as I know (@ReubenBond please correct me on this). We may end up with the continuations from tasks 1 and 2 accessing the same session concurrently.

I'm leaning in favour of isolated/pooled sessions because they mimic the behavior of well-known expensive resources such as `SqlConnection`, which product developers already tend to *handle with care*. Also, thread-safety and performance don't tend to mix, not sure how much concurrency handling would affect session performance, which already exists to *avoid* concurrency in the first place. What is your opinion?

---

I attempted to apply the `TaskCompletionSource` trick recommended by @ReubenBond on my PR clone, but wasn't able to do it on the "async lookup shell" (for lack of a better name) without it requiring changing the underlying lookup code and without avoiding allocation of `TaskCompletionSource` instances upfront.

Maybe you two have a better trick for this that I can learn from?
I'll try again on this new code base over the weekend and some good coffee.

---

> Alternatively, each grain could rent a session at activation and return it at deactivation. That would require something to pulse Refresh(), which we could potentially add hooks for (or use a timer).

This matches the intent of the sample and naturally follows into creating generic grain templates for any type, or with an eye in the future, providing an injected faster lookup for some type and underlying storage via DI bindings, azure function style.

ReubenBond:
>  Orleans is opinionated on this and enforces responsiveness in the system by killing any task taking over 30s 

One minor point: we wont kill Tasks, since .NET Tasks are co-operative & aren't preemptible

> This assumes developers won't pass around sessions between threads,

Just to be precise here: a grain method call can hop between threads between `await` points, so the session wouldn't strictly be thread-local

> When either continuation resumes, there's no guarantee the work will be performed on the same thread, as far as I know

Yes, precisely correct. This goes for any async method in .NET (i.e, it's not Orleans-specific)

JorgeCandeias:
> One minor point: we wont kill Tasks, since .NET Tasks are co-operative & aren't preemptible. We will throw a timeout exception back to the caller, though, so that they aren't left hanging forever. We will also destroy the grain activation if it has become unresponsive for a very long time.

Thanks for this important clarification 👍 

> 
> > This assumes developers won't pass around sessions between threads,
> 
> Just to be precise here: a grain method call can hop between threads between `await` points, so the session wouldn't strictly be thread-local

What I meant was it assumes developers won't *willingly* pass sessions around in a way that allows concurrent access.

badrishc:
> I attempted to apply the TaskCompletionSource trick recommended by @ReubenBond on my PR clone, but wasn't able to do it on the "async lookup shell" (for lack of a better name) without it requiring changing the underlying lookup code and without avoiding allocation of TaskCompletionSource instances upfront.

How about this: we slightly modify FASTER's API to support "ref Context" instead of "Context". User then passes in a "null" context to the lookup. If/when an operation goes async, it can fill up/augment this context via a new user-specified context provider callback. In our use case, this callback will allocate the `TaskCompletionSource` instance. It would be "ref" instead of "out" because users might still want to add stuff to the context (from the stack) before invoking the operation.

```cs
void UpdatePendingContext(ref context);
{
   context = new TaskCompletionSource<TOutput>(...);
}
```

JorgeCandeias:
> How about this: we slightly modify FASTER's API to support "ref Context" instead of "Context".

That's what I did and that's what I was trying to avoid.

JorgeCandeias:
I've updated [my clone with an alternate approach](https://github.com/JorgeCandeias/FASTER/blob/async-support/cs/src/core/Index/FASTER/FASTERThread.cs). Instead of relying on user code to set a ref variable on a callback (that may or may not happen), the `CompletePending` action is now itself async. This propagates down to the callbacks themselves, which, if called, must return at least an empty ValueTask (in order to implement the `IFunctions` interface, so it's now forced). If not called, that's okay too. The idea is to allow the user to continue pushing many data items to the lookup at once, while resolving any possible I/O at the end, the same way regular faster does it.

The `UpsertCompletionCallback` is updated as an example, as `ValueTask UpsertCompletionCallback(ref Key key, ref Value value, Context ctx)`.
However, this required adjusting the existing signature, so it's a breaking change.

Also, I haven't yet dealt with the blocking collection call regarding IO itself.

```csharp
internal void CompleteIOPendingRequests(FasterExecutionContext context)
{
    if (context.readyResponses.Count == 0) return;

    while (context.readyResponses.TryTake(out AsyncIOContext<Key, Value> request))
    {
        InternalContinuePendingRequestAndCallback(context, request);
    }
}
```

Do you think this approach is a viable option to replace that spinning for I/O with an async api?

badrishc:
Thanks @JorgeCandeias for your comments and explanation. It all makes sense.

In order to support a large number of sessions, I have redone the sessions interface so that it does not hold on to an entry in the epoch table. We ideally do not want the epoch table to grow larger than the number of cores/threads because we have to scan the epoch table in some situations.

In the current prototype, I let users create many "client sessions", and the client session simply copies the session context to the FASTER thread-local session on whatever thread the client session is currently running on. Client sessions are effectively async-local. Client sessions can scale up beyond the number of (internal) FASTER epoch entries which are thread-local. I still need to figure out how to integrate best with checkpointing, since we would need to complete the pending IO operations of sessions that are inactive as well.

I have added support for "ref Context" as an addition API, keeping the non-ref version for compatibility.

The async CompletePending is a good idea, but I don't see why the completion callbacks need to be async -- the callbacks are user code where there will likely not be async stuff going on. E.g., they just copy the value to an output variable. Instead, we can make the polling for I/O completion async by using an async semaphore.

badrishc:
After some hacking, I think it makes more sense, as you mentioned, to only have an async version of CompletePending, and not try to make individual operations async. This fits well in spirit with the notion of introducing batching in a non-invasive manner.

I am currently working on a truly async version of CompletePending (no spin wait or blocking) that can await all ops issued until that point in time. (This may even later be extended if needed to await all ops until a given seq. no. on that session.)

So, we don't need to add a ref context with this solution.

Checkpoint will also work similarly with an async CompleteCheckpoint, although that requires a bit more thinking/work internal to FASTER in order to notify client sessions, which may or may not be currently active.

badrishc:
Any idea why adding the NuGet package for ValueTask is causing C# Linux build on Mono to fail? https://dev.azure.com/ms/FASTER/_build/results?buildId=28238

ReubenBond:
The stack trace shows a MissingMethodException. I wonder if some tooling version needs updating

badrishc:
Yeah, updating .net core to latest fixed the issue.

badrishc:
@JorgeCandeias @ReubenBond et al.: we now have a prototype sessions API with async support. Please take it for a spin and let me know what you think.

Examples are here:
* https://github.com/microsoft/FASTER/blob/async-support/cs/test/SessionFASTERTests.cs
* https://github.com/microsoft/FASTER/blob/async-support/cs/test/AsyncLargeObjectTests.cs
* https://github.com/microsoft/FASTER/blob/async-support/cs/benchmark/FasterYcsbBenchmark.cs (this one uses a faster non-async version, but still using Session object)

Interestingly, benchmark perf is unaffected by these changes! For the non-async sessions (used only in FasterYcsbBenchmark), we get 180M ops/sec, similar to before.

For async sessions (the default), as long as we are issuing non-async ops to keys in memory, we get 160M ops/sec. This is in spite of doing *fine-grained* per-operation epoch resume and suspend! Users now do not need to call `Refresh()` periodically with the async sessions interface, which is the default. So it is more natural to C# developers. Feedback requested!

```cs
            using (var session = fht.StartClientSession())
            {
                InputStruct input = default(InputStruct);
                OutputStruct output = default(OutputStruct);

                var key1 = new KeyStruct { kfield1 = 13, kfield2 = 14 };
                var value = new ValueStruct { vfield1 = 23, vfield2 = 24 };

                session.Upsert(ref key1, ref value, Empty.Default, 0);
                var status = session.Read(ref key1, ref input, ref output, Empty.Default, 0);

                if (status == Status.PENDING)
                {
                    await session.CompletePendingAsync();
                }
                else
                {
                    Assert.IsTrue(status == Status.OK);
                }

                Assert.IsTrue(output.value.vfield1 == value.vfield1);
                Assert.IsTrue(output.value.vfield2 == value.vfield2);
            }
```

ReubenBond:
Great! For write operations, will subsequent status be pending almost always (unless some background flush occurred)?

I wonder if the operation should return ValueTask<T> instead of Status. If the status is Ok then the result is completed synchronously and if it's not Ok then the result must be awaited. It doesn't work with out/ref params, so the method would return ValueTask<T> where T is some tuple or result type, but I believe it's more natural

badrishc:
Upsert operations will almost always complete synchronously. Read operations might return a status of pending if they are reading something that is older and not in memory.

The idea of individual operations returning a Status is that in the FASTER session model, user session can issue batches of requests synchronously at extremely high performance, and periodically compete all (hopefully rare) outstanding pending operations by calling async CompletePending, or calling sync CompletePending to complete only the pending ops that have already returned from IO.

Making every read etc. operation return ValueTask will be a breaking change, and will require us to break the current invariant that all work happens on the user thread/session, never in an async thread (expect for the actual IO completion, which just queues the competition to a session local queue for later pull-based consumption by the session (via the CompletePending call). Also, the number of parked tasks will grow to the number of outstanding operations, instead of just one task for CompletePending.

Also, see these comments:
* https://github.com/microsoft/FASTER/pull/130#issuecomment-515549467
* https://github.com/microsoft/FASTER/pull/130#issuecomment-517769617

cs/src/core/ClientSession/ClientSession.cs
```diff
@@ -0,0 +1,198 @@
...
+        public void Dispose()
+        {
+            CompletePending(true);
+            fht.DisposeClientSession(ID);
+        }
+
+        /// <summary>
+        /// Resume session on current thread
+        /// Call SuspendThread before any async op
+        /// </summary>
```
gunaprsd: Would be nice to say how it is unsafe here

cs/src/core/Index/FASTER/FASTERThread.cs
```diff
@@ -52,129 +54,155 @@ internal long InternalContinue(Guid guid)
...
+                ctx.ioPendingRequests = new Dictionary<long, PendingContext>();
             }
+        }
 
-            threadCtx.Value = ctx;
+        internal void CopyContext(FasterExecutionContext src, FasterExecutionContext dst)
+        {
+            dst.phase = src.phase;
+            dst.version = src.version;
+            dst.markers = src.markers;
```
gunaprsd: Not sure if this array assignment is right. Don't we want a fresh set of markers for the new context? You seem to be invoking `CopyContext` within `AtomicSwitch` if the dst.version is smaller than current FASTER version. Is it okay to do a shallow copy of array in that case?

cs/src/core/Index/Recovery/Checkpoint.cs
```diff
@@ -673,7 +497,7 @@ private SystemState GetNextState(SystemState start, CheckpointType type = Checkp
                     nextState.version = start.version + 1;
                     break;
                 case Phase.IN_PROGRESS:
-                    nextState.phase = Phase.WAIT_PENDING;
+                    nextState.phase = RelaxedCPR ? Phase.WAIT_FLUSH : Phase.WAIT_PENDING;
```
gunaprsd: Please update the comment above the `GetNextState` and `GlobalMoveToNextState` functions to reflect the new state machine.

cs/src/core/ClientSession/FASTERAsync.cs
```diff
@@ -0,0 +1,365 @@
...
+                {
+                    currentState.word = previousState.word;
+                }
+                else if (previousState.version < finalState.version)
+                {
+                    if (finalState.phase <= Phase.PREPARE)
+                        previousState.version = finalState.version;
+                    else
+                        previousState.version = finalState.version - 1;
+                    currentState = GetNextState(previousState, CheckpointType.FULL);
```
gunaprsd: Does it always have to be a FULL checkpoint for this to happen?

cs/src/core/ClientSession/FASTERAsync.cs
```diff
@@ -0,0 +1,365 @@
...
+                                ctx.markers[EpochPhaseIdx.PrepareForIndexCheckpt] = false;
+                            }
+
+                            if (async && !IsIndexFuzzyCheckpointCompleted())
+                            {
+                                clientSession.UnsafeSuspendThread();
+                                await IsIndexFuzzyCheckpointCompletedAsync();
+                                clientSession.UnsafeResumeThread();
+                            }
+                            GlobalMoveToNextCheckpointState(currentState);
```
gunaprsd: Don't we need to check if `IsIndexFuzzyCheckpointCompleted()` for the non-async case as well?

cs/src/core/ClientSession/FASTERAsync.cs
```diff
@@ -0,0 +1,365 @@
...
+                                if (FoldOverSnapshot)
+                                {
+                                    notify = (hlog.FlushedUntilAddress >= _hybridLogCheckpoint.info.finalLogicalAddress);
+                                }
+                                else
+                                {
+                                    notify = (_hybridLogCheckpoint.flushedSemaphore != null) && _hybridLogCheckpoint.flushedSemaphore.CurrentCount > 0;
+                                }
+
+                                if (async && !notify)
```
gunaprsd: Can you write a comment on why exactly you are doing within `async && !notify` ?

badrishc:
We now have a full async-compliant sessions interface to FasterKv! You no longer need to call refresh or worry about thread affinity. The overall idea is to create new sessions to FasterKv and perform a sequence of operations on a session. You can also have a separate async committer (checkpointer) thread for durability. The async session operations can be configured to await until either after they are completed in memory to made durable by a checkpoint. This is similar to the new FasterLog API.

See any example in the playground for details (older API is marked obsolete for now). A good one is the newly added sample [here](https://github.com/microsoft/FASTER/blob/0237a0c439d9ed1349114911db9800bed8be051e/cs/playground/FasterKVAsyncSample/Program.cs).

I think it almost cannot get easier than this for C# users, but would love to hear feedback from everyone.

badrishc:
## Async Sessions API

```cs
// Create new shared FASTER KV
var faster = new FasterKV(...);

// Create any number of sessions to FasterKV
var session = faster.NewSession();

// Async read, retrieves from disk if needed, reads uncommited if present
(Status, Output) result = await session.ReadAsync(key);

// Read waits for uncommitted read value to checkpoint/commit before returning
(Status, Output) result = await session.ReadAsync(key, waitForCommit: true);

// Upsert into mutable region, does not wait for commit
await session.UpsertAsync(key, value);

// Upsert into mutable region, return after this (and all prev) upserts commit
await session.UpsertAsync(key, value, waitForCommit: true);

// RMW into mutable region, does not wait for commit
await session.RMWAsync(key, input);

// RMW into mutable region, wait for commit
await session.RMWAsync(key, input, waitForCommit: true);

// Waits for all prev operations on session to commit before returning
await session.WaitForCommit();
```

Commits can be performed periodically by a separate thread/task:
```cs
int i = 0;
while (true)
{
  Thread.Sleep(5000);
  if (i++ % 100 == 0)
     faster.TakeFullCheckpoint(out _); // periodically take full index + log checkpoint
  else
     faster.TakeHybridLogCheckpoint(out _); // take (incremental) log checkpoint
}
```

cs/src/core/Index/Common/Contexts.cs
```diff
@@ -148,20 +161,16 @@ public struct HybridLogRecoveryInfo
...
-        /// Guid array
-        /// </summary>
-        public Guid[] guids;
 
         /// <summary>
-        /// Tokens per guid restored during Continue
+        /// Commit tokens per session restored during Continue
         /// </summary>
-        public ConcurrentDictionary<Guid, long> continueTokens;
+        public ConcurrentDictionary<string, CommitPoint> continueTokens;
```
tli2: Is using a string instead of Guid a performance issue?

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -23,8 +26,16 @@ public unsafe partial class FasterKV<Key, Value, Input, Output, Context, Functio
...
         private readonly bool WriteDefaultOnDelete = false;
+        private bool RelaxedCPR = false;
+
+        /// <summary>
+        /// Use relaxed version of CPR, where ops pending I/O
+        /// are not part of CPR checkpoint. This mode allows
+        /// us to eliminate the WAIT_PENDING phase, and allows
+        /// sessions to be suspended. Do not modify during checkpointing.
+        /// </summary>
+        public void UseRelaxedCPR() => RelaxedCPR = true;
```
tli2: Is there a way for us to ensure that this is not modified during checkpointing?

Will people want to modify this at all for a single Faster instance?

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -186,67 +186,67 @@ public FasterKV(long size, Functions functions, LogSettings logSettings, Checkpo
...
         }
 
         /// <summary>
-        /// Take full checkpoint
+        /// Initiate full checkpoint
         /// </summary>
-        /// <param name="token"></param>
-        /// <returns></returns>
+        /// <param name="token">Checkpoint token</param>
+        /// <returns>Whether we could initiate the checkpoint</returns>
```
tli2: nit: maybe it's helpful to also provide why a checkpoint might fail to be initiated here

cs/src/core/Index/Interfaces/IFasterKV.cs
```diff
@@ -103,8 +100,37 @@ public interface IFasterKV<Key, Value, Input, Output, Context> : IDisposable
...
+        /// Session starts in dormant state.
+        /// </summary>
+        /// <param name="sessionId"></param>
+        /// <param name="supportAsync"></param>
+        /// <returns></returns>
+        ClientSession<Key, Value, Input, Output, Context, Functions> NewSession(string sessionId = null, bool supportAsync = true);
 
-        /* Recovery */
+        /// <summary>
+        /// Continue client session with FASTER
```
tli2: nit: it's probably better to clarify whether this is used exclusively for recovery from failure, or some how a session can be "suspended" as part of normal operation

cs/src/core/Index/Common/Contexts.cs
badrishc: Might be okay because it is not accessed in critical paths. Strings make it easier to create and reference meaningful session names. e.g. user can create sessions in code such as "s1", "s2", etc. so they dont need to persist it. If they don't provide a session name, we auto name it to a Guid converted to string. The downside is that on every commit, we have to write each session name string to disk as part of commit.

PR #133, opened 2019-06-11
Support for callback when records in hybrid log become read-only

badrishc:
This PR allows users to register a user-defined callback, which is invoked when records in the hybrid log become read-only. It can be used for various reasons:

(1) to perform data write-through to an external store
(2) to perform stream processing on the change log created by the key-value store

How it Works
---------------

The FASTER log accessor now supports the `IObservable<IFasterScanIterator<Key, Value>>` interface. This means you can `Subscribe` to FASTER log, in order to receive (as push calls) updates whenever records enter the read-only region of the hybrid log:

```cs
        var s = fht.Log.Subscribe(new LogObserver());
        ...
        s.Dispose();
```

The parameter `LogObserver` is an instance of type `IObserver<IFasterScanIterator<Key, Value>>`. User gets push-based callbacks via the `OnNext` method of the observer. Each callback provides an iterator for a batch of records that have just become read-only, in bulk. The user can then perform operations such as stream processing or write-through to the remote database.

Here is an example of the callback:

```cs
        class LogObserver : IObserver<IFasterScanIterator<Key, Value>>
        {
            public void OnCompleted()
            {
            }

            public void OnError(Exception error)
            {
            }

            public void OnNext(IFasterScanIterator<Key, Value> iter)
            {
                while (iter.GetNext(out RecordInfo info, out Key key, out Value value))
                {
                    // Process key and value
                }
                iter.Dispose();
            }
        }
```

### DIFF
--- a/cs/src/core/Allocator/AllocatorBase.cs
+++ b/cs/src/core/Allocator/AllocatorBase.cs
@@ -220,6 +220,11 @@ namespace FASTER.core
         /// </summary>
         protected readonly Action<long, long> EvictCallback = null;
 
+        /// <summary>
+        /// Observer for records entering read-only region
+        /// </summary>
+        internal IObserver<IFasterScanIterator<Key, Value>> OnReadOnlyObserver;
+
         #region Abstract methods
         /// <summary>
         /// Initialize
@@ -597,6 +602,8 @@ namespace FASTER.core
             if (ownedEpoch)
                 epoch.Dispose();
             bufferPool.Free();
+
+            OnReadOnlyObserver?.OnCompleted();
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


badrishc:
We have also added an extension method to convert an `IObservable<IFasterScanIterator<Key, Value>>` to an easier-to-use `IObservable<Record<Key, Value>>`. This converter copies the key and value from the log to the `Record<Key, Value>` struct, so it will not directly work with special inline variable-length structs (see https://github.com/microsoft/FASTER/pull/120).

Usage:
```cs
        var s = fht.Log.ToRecordObservable.Subscribe(...);
        ...
        s.Dispose();
```

badrishc:
Currently, we support only one subscriber to the log (easy to extend in future). Also, note that the subscriber only receives new log updates from the time of subscription onwards. In other words, no I/O is associated with this operation. If you need to scan the historical part of the log, use the Scan(...) method instead (see https://github.com/microsoft/FASTER/pull/90).

PR #145, opened 2019-06-22
[c++] Add a new `value_size()` method to `RmwContext` for RCU operations

### LINKED ISSUE #136: [c++] Resize Value allocation during `RmwCopy`
### Motivation
We would like to be able to store variable length values in FASTER and use RMW to make them (potentially) bigger. As a motivating example we would like to store a String and then use RMW to modify the String based on the previous value .

### Background
We have a set up for supporting variable length values similar to that used in these tests: 
* https://github.com/microsoft/FASTER/blob/master/cc/test/in_memory_test.cc#L1319
* https://github.com/microsoft/FASTER/blob/master/cc/test/in_memory_test.cc#L596

Our `RmwContext` will call a function (defined elsewhere) with the old String to get the new String.
```c++
  class RmwContext : public IAsyncContext {
   public:
    typedef Key key_t;
    typedef Value value_t;

    RmwContext(uint64_t key)
      : key_{ key } {
    }

    /// Copy (and deep-copy) constructor.
    RmwContext(const RmwContext& other)
      : key_{ other.key_ } {
    }

    /// The implicit and explicit interfaces require a key() accessor.
    inline const Key& key() const {
      return key_;
    }
    inline uint32_t value_size() const {
      return sizeof(value_t); // This is where insufficient space is allocated
    }

    inline void RmwInitial(Value& value) {
      char* new_string = get_new_string("");
      value.gen_lock_.store(GenLock{});
      value.size_ = sizeof(Value) + sizeof(new_string);
      value.length_ = sizeof(new_string);
      std::memcpy(value.buffer(), new_string, sizeof(new_string));
    }
    inline void RmwCopy(const Value& old_value, Value& value) {
      char* new_string = get_new_string(old_value.buffer());
      value.gen_lock_.store(GenLock{});
      value.size_ = sizeof(Value) + sizeof(new_string);
      value.length_ = sizeof(new_string);
      std::memcpy(value.buffer(), new_string, sizeof(new_string));
    }
    inline bool RmwAtomic(Value& value) {
      bool replaced;
      while(!value.gen_lock_.try_lock(replaced) && !replaced) {
        std::this_thread::yield();
      }
      if(replaced) {
        // Some other thread replaced this record.
        return false;
      }
      char* new_string = get_new_string(value.buffer());
      if(value.size_ < sizeof(Value) + sizeof(new_string)) {
        // Current value is too small for in-place update.
        value.gen_lock_.unlock(true);
        return false;
      }
      // In-place update overwrites length and buffer, but not size.
      value.length_ = sizeof(new_string);
      memcpy(value.buffer(), new_string, sizeof(new_string));
      value.gen_lock_.unlock(false);
      return true;
    }

   protected:
    /// The explicit interface requires a DeepCopy_Internal() implementation.
    Status DeepCopy_Internal(IAsyncContext*& context_copy) {
      return IAsyncContext::DeepCopy_Internal(*this, context_copy);
    }

   private:
    Key key_;
};
```

### Problem
This way of trying to use RMW to modify a String does not work because insufficient space is allocated for the new record. https://github.com/microsoft/FASTER/blob/master/cc/src/core/faster.h#L1076 shows that the new record will be given `RmwContext::value_size()` bytes for the new Value. In the case of a Value like in this example, we would need to know the size of `new_string` in order to know how much space to reserve for the new Value.

### My Question
Is there a way in which we can resize a Value's allocation on the `HybridLog` during an RMW operation? With this we could extend the buffer of the new record to fit `new_string`. To be clear, I'm asking about resizing a record as it is being first written to, not a record already fully-created.

We could of course do a Read request then construct the new String and do an Upsert (given we now would know the required size to allocate) but this would be missing out on the ability to do an RMW in one operation. I guess it would still be possible to take advantage of in-place update during the Upsert?

matthewbrookes:
What:
* add a `value_size()` method that allows the user to compute the new Value size with respect to the old Value
* update tests etc to include the new method in `RmwContext`
* add a test to show the use of this method for a growing string

Why:
* supports RMW operations for a "growing" type such as a string, vector

Closes #136

### DIFF
--- a/cc/benchmark-dir/benchmark.cc
+++ b/cc/benchmark-dir/benchmark.cc
@@ -222,6 +222,9 @@ class RmwContext : public IAsyncContext {
   inline static constexpr uint32_t value_size() {
     return sizeof(value_t);
   }
+  inline static constexpr uint32_t value_size(const value_t& old_value) {
+    return sizeof(value_t);
+  }
 
   /// Initial, non-atomic, and atomic RMW methods.
   inline void RmwInitial(value_t& value) {
--- a/cc/playground/sum_store-dir/sum_store.h
+++ b/cc/playground/sum_store-dir/sum_store.h
@@ -102,6 +102,9 @@ class RmwContext : public IAsyncContext {
   inline static constexpr uint32_t value_size() {
     return sizeof(value_t);
   }
+  inline static constexpr uint32_t value_size(const NumClicks& old_value) {
+    return sizeof(value_t);
+  }
 
  protected:
   /// The explicit interface requires a DeepCopy_Internal() implementation.
--- a/cc/src/core/faster.h
+++ b/cc/src/core/faster.h
@@ -1074,7 +1074,14 @@ inline OperationStatus FasterKv<K, V, D>::InternalRmw(C& pending_context, bool r
 
   // Create a record and attempt RCU.
 create_record:
-  uint32_t record_size = record_t::size(key, pending_context.value_size());
+  uint32_t record_size;
+  const record_t* old_record;
+  if (address >= head_address) {
+    old_record = reinterpret_cast<const record_t*>(hlog.Get(address));
+    record_size = record_t::size(key, pending_context.value_size(old_record));
+  } else {
+    record_size = record_t::size(key, pending_context.value_size());
+  }
   Address new_address = BlockAllocate(record_size);
   record_t* new_record = reinterpret_cast<record_t*>(hlog.Get(new_address));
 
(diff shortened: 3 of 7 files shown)


badrishc:
Thanks, this looks reasonable. Since it's a core change we will need a bit more testing so might take some time to merge. Have you done any performance regression tests with the benchmarks?

matthewbrookes:
I haven't yet but I may get a chance over the weekend to run the RMW benchmark and compare

matthewbrookes:
Hey, I ran the RMW benchmark against `master` and against this new branch. You can see my results in the following spreadsheet.

https://docs.google.com/spreadsheets/d/1NozTEkdRO1RHmjFBRhT3oxQExCy3TlXPTheyvYBpdhA/edit?usp=sharing

Whilst the results suggest an improvement in performance with these modifications, I think the difference isn't statistically significant and suggests that this change would not impact performance.

PR #147, opened 2019-06-28
Page blob device

### LINKED ISSUE #139: Storage device for Azure Page Blobs
[Azure Page Blobs](https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs) are made up of 512-byte pages up to 8 TB in total size and are designed for frequent random read/write operations, making them a good candidate for serving as a storage device for FASTER. We would like to implement the `IDevice` abstraction for Azure Page Blobs, along the lines of existing devices such as `LocalStorageDevice`.

tli2:
This PR implements an `IDevice` backed by Azure page blob storage.

Still TODO:

-  ~~So far this PR is tested by replacing  `Devices::CreateLogDevice` with one that returns `AzurePageBlobStorage` instead of `LocalStorageDevice`. All tests pass, but performance is an order of magnitude slower. So probably should not check those in.~~

- ~~For unit tests, we will want to use an emulator. Need to figure out how to set that up with Azure pipelines.~~
- ~~Unit test performance is visibly worse than local disk when using emulator, and even worse than that when using a real storage account. While it is expected, I need to run some more benchmarks to see if the slowdown is reasonable.~~

Fix #139


tli2:
Testing of this PR also hindered by #148, although not completely blocked,

tli2:
Azure Storage Emulator does not run on Linux. Will investigate [alternative emulator](https://github.com/Azure/Azurite) for use in unit tests

badrishc:
I think this device should be in its own project instead of within FASTER.core. Then, users can opt in to the feature, otherwise they would have to take a NuGet dependency on the Azure related NuGet just to run FASTER core. Project and namespace could be named FASTER.devices.AzurePageBlobDevice.

cs/src/core/Device/AzurePageBlobDevice.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <summary>
+        /// <see cref="IDevice.ReadAsync(int, ulong, IntPtr, uint, IOCompletionCallback, IAsyncResult)">Inherited</see>
+        /// </summary>
+        public override unsafe void ReadAsync(int segmentId, ulong sourceAddress, IntPtr destinationAddress, uint readLength, IOCompletionCallback callback, IAsyncResult asyncResult)
+        {
+            CloudPageBlob pageBlob = GetOrAddPageBlob(segmentId);
+
+            // Even though Azure Page Blob does not make use of Overlapped, we populate one to conform to the callback API
+            Overlapped ov = new Overlapped(0, 0, IntPtr.Zero, asyncResult);
+            NativeOverlapped* ovNative = ov.UnsafePack(callback, IntPtr.Zero);
```
peterfreiling: this will leak unless the callback is actually called (e.g., due to exceptions, etc.). Probably need to wrap with try/finally.

cs/src/core/Device/AzurePageBlobDevice.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            UnmanagedMemoryStream stream = new UnmanagedMemoryStream((byte*)destinationAddress, readLength, readLength, FileAccess.Write);
+
+            // TODO(Tianyu): This implementation seems to swallow exceptions that would otherwise be thrown from the synchronous version of this
+            // function. I wasn't able to find any good documentaiton on how exceptions are propagated or handled in this scenario. 
+            pageBlob.BeginDownloadRangeToStream(stream, (Int64)sourceAddress, readLength, ar => callback(0, readLength, ovNative), asyncResult);
```
peterfreiling: >ar => callback(0, readLength, ovNative) [](start = 90, length = 39)

What are the guarantees of this lambda? We need to ensure that the callback is invoked exactly once, and if it's not invoked at all, free the overlapped, and invoke callback with the appropriate error.

cs/src/core/Device/AzurePageBlobDevice.cs
```diff
@@ -0,0 +1,136 @@
...
+            // Even though Azure Page Blob does not make use of Overlapped, we populate one to conform to the callback API
+            Overlapped ov = new Overlapped(0, 0, IntPtr.Zero, asyncResult);
+            NativeOverlapped* ovNative = ov.UnsafePack(callback, IntPtr.Zero);
+            UnmanagedMemoryStream stream = new UnmanagedMemoryStream((byte*)sourceAddress, numBytesToWrite);
+            pageBlob.BeginWritePages(stream, (long)destinationAddress, null, ar => callback(0, numBytesToWrite, ovNative), asyncResult);
+        }
+
+        private CloudPageBlob GetOrAddPageBlob(int segmentId)
+        {
+            return blobs.GetOrAdd(segmentId, id => CreatePageBlob(id));
```
peterfreiling: >id => CreatePageBlob(id) [](start = 45, length = 24)

This factory is not guaranteed to be called exactly once - will we leak blobs in this case? Maybe use Lazy to guarantee exactly-once instantiation?

cs/src/core/Device/AzurePageBlobDevice.cs
tli2: I have been unable to find any documentation on what exceptions are thrown from the Azure call. However, if one is thrown, it will be unhandled and the application will be halted, and memory leak in this situation is a non-issue.

Throwing an exception seems reasonable because there is little the code can do to recover at the Device level.

peterfreiling: Unfortunately unhandled exceptions from thread pool threads do not crash the app, and instead terminate the background thread and fail silently. It is also a common pattern to retry a failed operation, especially for cloud operations that suffer from intermittent failures. But aside from memory leak, I was more worried about the blob leak - I want to make sure we don't create blobs in the remote storage that are never cleaned up.

---
In reply to: [303554075](https://github.com/microsoft/FASTER/pull/147#discussion_r303554075) [](ancestors = 303554075)

cs/src/cloud/AzurePageBlobDevice.cs
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    pageBlob.BeginCreate(size, ar => {
+                        pageBlob.EndCreate(ar);
+                        WriteToBlobAsync(pageBlob, sourceAddress, destinationAddress, numBytesToWrite, callback, asyncResult);
+                    }, null);
+                }
+                else
+                {
+                    // Some other thread beat us to calling create, should use their handle to invoke write directly instead
+                    WriteToBlobAsync(blobs[segmentId], sourceAddress, destinationAddress, numBytesToWrite, callback, asyncResult);
```
peterfreiling: Do we need to somehow delete the page blob reference in this case?

tli2: I believe it is just a stateless stub that is GCed as references to it are dropped. Again, no way to be sure since I have not found any documentation or specifications stating this.

tli2:
@badrishc I had to add a lot more code to deal with the race around Create.

Will squash all these commits when merging.

Issue #149, opened 2019-06-30
Memory reorder in Protect, ProtectAndDrain and Unprotect.


Transpeptidase:
In function ``Protect``, 

> https://github.com/microsoft/FASTER/blob/master/cc/src/core/light_epoch.h

``` cpp
inline uint64_t Protect() {
    uint32_t entry = Thread::id();
    table_[entry].local_current_epoch = current_epoch.load();
    return table_[entry].local_current_epoch;
  }
```
Because of CPU memory consistency model, the store of  `` table_[entry].local_current_epoch``  may be reordered after some protected code sections. Then, the ``ComputeNewSafeToReclaimEpoch`` function will get the wrong ``safe_to_reclaim_epoch``.

tli2:
Do you have a trace that triggers this problem or did you find it through inspection?

Transpeptidase:
```
-------------- thread T1 ----------------------
| epoch 2   local_current_epoch == kUnprotected
| epoch 2   tmp = current_epoch.load() = 2              
| epoch 2   read shared memory A
| epoch 3   read shared memory A                      --------------> thread T2 Drains
| epoch 3   local_current_epoch = tmp = 2                            
| epoch 3   local_current_epoch = kUnprotected      
| epoch 3                                                                 
```

In the above example,  "local_current_epoch = tmp = 2" can be reordered after reading shared memory.
When thread T2 Drains,  it will calculate ``safe_to_reclaim_epoch`` equal to 2. But epoch 2 is not safe (thread T1 is reading shared memory A)

tli2:
I was more asking if you have a test case or workload you used to reliably reproduce this? That would be helpful in verifying if the bug is fixed.

Transpeptidase:
[test.txt](https://github.com/microsoft/FASTER/files/3355333/test.txt)

This is test source code. 
Run ``g++ -std=c++14 -pthread test.cc thread.cc lss_allocator.cc  -O3`` under directory ``core``.

To make the bug easy to reproduce， insert ``clflush`` in the ``Protect`` (make store of ``local_current_epoch`` easy to be reordered)

``` cpp
 inline uint64_t Protect() {
  
    uint32_t entry = Thread::id();
  
    asm volatile("clflush (%0)" ::"r"(&table_[entry].local_current_epoch));
  
    table_[entry].local_current_epoch = current_epoch.load();
    return table_[entry].local_current_epoch;
  }   

```

Transpeptidase:
[...]

PR #150, opened 2019-07-04
epoch fix

Transpeptidase:
This PR fixes a bug in LightEpoch, which caused by memory reorder.

Fix [#149](https://github.com/microsoft/FASTER/issues/149)


badrishc:
Seems to be failing a few unit tests

Transpeptidase:
> Seems to be failing a few unit tests

I don't know if it is running too slowly or deadlock.

Transpeptidase:
> Seems to be failing a few unit tests

I fixed it.

cc/src/core/light_epoch.h
```diff
@@ -176,44 +176,58 @@ class LightEpoch {
...
+
+    return e;
   }
 
   /// Enter the thread into the protected code region
   /// Process entries in drain list if possible
   inline uint64_t ProtectAndDrain() {
     uint32_t entry = Thread::id();
-    table_[entry].local_current_epoch = current_epoch.load();
+    auto e = current_epoch.load(std::memory_order_relaxed);
```
tli2: These three lines of code appears duplicated multiple times throughout the file. Consider putting them into a function?

Transpeptidase:
> First of all, thanks for finding the bug and submitting a PR!
> 
> If you have more time to work on this:
> 
> * The use of std::atomic mixed with fences seems overly complicated. By default every operation on atomic should operate under `std::memory_order_seq_cst` and should do the correct thing for you. Now it is conceivable that we lose some performance due to excessive synchronization, but let's back that up with benchmark results and not prematurely optimize.
> * Alternatively, consider using `std::atomic_signal_fence` without the need to turn `local_current_epoch` into an atomic.
> * Can you add your test code as a unit test?

Unfortunately, in C++ consistency model, ``memory_order_acquire`` only prevents the reorder of load-store and load-load. Thus, the ``store(memory_order_acquire)`` and ``std::atomic_thread_fence
(memory_order_acquire)`` don't work. ``memory_order_seq_cst`` is not always doing  the correct thing,  the ``store(memory_order_seq_cst)`` is equivalent to ``store(memory_order_acquire)``, which 
allows the reorder of  store-load. (see https://en.cppreference.com/w/cpp/atomic/memory_order,  and you can verify it  by generate assembly code with ``g++ -O3 -S``). 
I have not used `std::atomic_signal_fence` before. But  it seems to only prevent reorder by compiler rather than reorder by CPU (https://en.cppreference.com/w/cpp/atomic/atomic_signal_fence).

tli2:
> First of all, thanks for finding the bug and submitting a PR!
> If you have more time to work on this:
> 
> The use of std::atomic mixed with fences seems overly complicated. By default every operation on atomic should operate under std::memory_order_seq_cst and should do the correct thing for you. Now it is conceivable that we lose some performance due to excessive synchronization, but let's back that up with benchmark results and not prematurely optimize.
> Alternatively, consider using std::atomic_signal_fence without the need to turn local_current_epoch into an atomic.
> Can you add your test code as a unit test?
> 
> 
> Unfortunately, in C++ consistency model, memory_order_acquire only prevents the reorder of load-store and load-load. Thus, the store(memory_order_acquire) and std::atomic_thread_fence (memory_order_acquire) don't work. memory_order_seq_cst is not always doing  the correct thing,  the store(memory_order_seq_cst) is equivalent to store(memory_order_acquire), which
> allows the reorder of  store-load. (see https://en.cppreference.com/w/cpp/atomic/memory_order,  and you can verify it  by generate assembly code with g++ -O3 -S).
> I have not used std::atomic_signal_fence before. But  it seems to only prevent reorder by compiler rather than reorder by CPU (https://en.cppreference.com/w/cpp/atomic/atomic_signal_fence).

Seems like it's worth a comment in the code in your PR.

badrishc:
We have not yet seen this to be an issue that repros in FASTER. The matter requires further investigation. I wonder if it happens only on certain architectures. Closing for now, we can reopen if we get a clearer idea of the issue in terms of a FASTER testcase failure.

PR #158, opened 2019-07-22
Object log recovery fixes

badrishc:
The logic for reading records in the presence of an object log was writing directly to a pinned array of managed objects, to avoid fine grained copy of parts of the log that are not managed. However, this solution has problems in the presence of GC. This checkin primarily undoes this optimization for the sake of correctness. There are a few other minor improvements, such as using a SafeConcurrentDictionary to store file handles, to avoid spurious creation when many threads read in parallel.


badrishc:
We also add an option to enable file buffering for the storage device, and ensure that during recovery, we only load records from HeadAddress onwards into main memory.

Issue #148, opened 2019-06-28
Mixed use of IDevice and direct local file access


tli2:
In the checkpointing code (`Checkpoint.cs, FasterKV::GlobalMoveToNextState`), logs are written through the `IDevice` interface by calling `Devices::CreateLogDevice`. However, on the recovery code path (`Contexts.cs, IndexRecoveryInfo::Recover`), the system is hard-coded to use `StreamReader` on the given file path, assuming that the file exists on the local disk.

This results in failure of recovery tests when one naively changes the implementation of `Devices::CreateLogDevice` to supply a non-file based device, but fails to special case checkpointing code. 

The failure to find recovery file also seems to be silent.

badrishc:
Closing as PR has merged.

PR #161, opened 2019-07-23
Make checkpointing pluggable

badrishc:
The goal of this PR is to make FASTER C# checkpointing use a pluggable user-specified interface for providing devices (such as for index and snapshot) and for performing the atomic metadata commit. We will refactor the current implementation (currently hard coded to use LocalStorageDevice and C# Streams) as a reference implementation.

This will allow users to provide plugins that might, for example, (1) write checkpoints to remote devices such as Azure page blobs; (2) use a DBMS to commit and store the checkpoint metadata; and (3) write custom checkpoint adapters that for example, may encrypt or add checksums to the checkpoint data/metadata.

### DIFF
--- a/cs/src/core/Index/Recovery/ICheckpointManager.cs
+++ b/cs/src/core/Index/Recovery/ICheckpointManager.cs
@@ -43,31 +43,31 @@ namespace FASTER.core
         /// Commit index checkpoint (synchronous)
         /// </summary>
         /// <param name="indexToken"></param>
-        /// <param name="commitInfo"></param>
+        /// <param name="commitMetadata"></param>
         /// <returns></returns>
-        void CommitIndexCheckpoint(Guid indexToken, byte[] commitInfo);
+        void CommitIndexCheckpoint(Guid indexToken, byte[] commitMetadata);
 
         /// <summary>
         /// Commit log checkpoint (synchronous)
         /// </summary>
         /// <param name="logToken"></param>
-        /// <param name="commitInfo"></param>
+        /// <param name="commitMetadata"></param>
         /// <returns></returns>
-        void CommitLogCheckpoint(Guid logToken, byte[] commitInfo);
+        void CommitLogCheckpoint(Guid logToken, byte[] commitMetadata);
 
         /// <summary>
         /// Retrieve commit info for previous index checkpoint
         /// </summary>
         /// <param name="indexToken"></param>
         /// <returns>Commit info, if valid checkpoint found, and null otherwise</returns>
-        byte[] GetIndexCommitInfo(Guid indexToken);
+        byte[] GetIndexCommitMetadata(Guid indexToken);
 
         /// <summary>
         /// Retrieve commit info for previous log checkpoint
         /// </summary>
         /// <param name="logToken"></param>
         /// <returns>Commit info, if valid checkpoint found, and null otherwise</returns>
-        byte[] GetLogCommitInfo(Guid logToken);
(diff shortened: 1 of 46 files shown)


badrishc:
Fixes https://github.com/microsoft/FASTER/issues/148

cs/src/core/Index/Recovery/ICheckpointManager.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <param name="indexToken"></param>
+        /// <returns>Commit info, if valid checkpoint found, and null otherwise</returns>
+        byte[] GetIndexCommitMetadata(Guid indexToken);
+
+        /// <summary>
+        /// Retrieve commit info for previous log checkpoint
+        /// </summary>
+        /// <param name="logToken"></param>
+        /// <returns>Commit info, if valid checkpoint found, and null otherwise</returns>
+        byte[] GetLogCommitMetadata(Guid logToken);
```
peterfreiling2: It's unclear from the interface whether this needs to be implemented for fold over checkpoints

cs/src/core/Index/Recovery/ICheckpointManager.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// Get latest valid checkpoint
+        /// </summary>
+        /// <param name="indexToken"></param>
+        /// <param name="logToken"></param>
+        /// <returns>true if latest valid checkpoint found, false otherwise</returns>
+        bool GetLatestCheckpoint(out Guid indexToken, out Guid logToken);
```
peterfreiling2: It's unclear from the interface how this works for fold over checkpoints

cs/src/core/Index/Recovery/ICheckpointManager.cs
badrishc: Yes, it needs to be implemented for all log checkpoints (fold-over and snapshot). In case of foldover, this has information related to the status of the log as of the checkpoint, so we can resume from the correct log location (e.g., to handle the case if more uncheckpointed activity occurred on the log before the crash).

cs/src/core/Index/Recovery/ICheckpointManager.cs
badrishc: All full checkpoints (foldover or snapshot does not matter) consist of a Guid for the index and a Guid for the log. The implementer of GetLatestCheckpoint simply returns these two Guids. For a "full checkpoint" (index+log) these two Guids will usually be the same.

badrishc:
FASTER calls the checkpoint manager interface during checkpoint/recovery in this sequence:

Checkpoint:
*  InitializeIndexCheckpoint (for index checkpoints) -> 
*  GetIndexDevice (for index checkpoints) ->
*  InitializeLogCheckpoint (for log checkpoints) ->
*  GetSnapshotLogDevice (for log checkpoints in snapshot mode) ->
*  GetSnapshotObjectLogDevice (for log checkpoints in snapshot mode with objects) ->
*  CommitLogCheckpoint (for log checkpoints) ->
*  CommitIndexCheckpoint (for index checkpoints) ->

Recovery:
*  GetLatestCheckpoint (if request to recover to latest checkpoint) ->
*  GetIndexCommitMetadata ->
*  GetLogCommitMetadata ->
*  GetIndexDevice ->
*  GetSnapshotLogDevice (for recovery in snapshot mode) ->
*  GetSnapshotObjectLogDevice (for recovery in snapshot mode with objects)

Provided devices will be closed directly by FASTER when done.

PR #163, opened 2019-08-06
[c++] in_memory_test: fix inverted rmw replace condition

matthewbrookes:
Currently, the example code has branches that do the opposite to what
the condition implies should happen. As a result, non-replaced records
are being marked as replaced and causing unnecessary read-copy-updates.

I discovered a bug in `faster-rs` where we were using the original logic and were performing subsequent RCU operations even though the values were being updated in-place.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,7 +27,7 @@ jobs:
   steps:
   - task: NuGetToolInstaller@0
     inputs:
-      versionSpec: '4.4.1' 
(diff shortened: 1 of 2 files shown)


matthewbrookes:
@badrishc please could re-run CI. I promise this isn't a breaking change for C# :laughing:

badrishc:
Haha, don't worry, I know exactly what the problem is with the break. Need to update a tool during the build. I'll take care of it. Thanks for the fix :).

PR #164, opened 2019-08-09
Implemented iterator for variable length structs.

marius-klimantavicius:
Current IFasterScanIterator interface is not suited for variable length structs as GetNext would only return a fragment of key/value (up to sizeof(Key)/sizeof(Value)). To solve this issue two new methods were added ref GetKey()/ref GetValue() that return references to actual key/value references in hlog.

There is disconnect in GenericScanIterator as it returns a reference to a copy of key/value and not direct reference to hlog. In case this is needed we could have flag to indicate whether we need to read from frame or hlog directly and store current page/offset.

This implementation does not support compaction for variable length structs.

### DIFF
--- a/cs/src/core/Allocator/BlittableScanIterator.cs
+++ b/cs/src/core/Allocator/BlittableScanIterator.cs
@@ -22,6 +22,7 @@ namespace FASTER.core
 
         private bool first = true;
         private long currentAddress, nextAddress;
+        private long currentPhysicalAddress;
 
         /// <summary>
         /// Current address
@@ -72,17 +73,31 @@ namespace FASTER.core
         }
 
         /// <summary>
-        /// Get next record in iterator
+        /// Gets reference to current key
+        /// </summary>
+        /// <returns></returns>
+        public ref Key GetKey()
+        {
+            return ref hlog.GetKey(currentPhysicalAddress);
+        }
+
+        /// <summary>
+        /// Gets reference to current value
+        /// </summary>
+        /// <returns></returns>
+        public ref Value GetValue()
+        {
+            return ref hlog.GetValue(currentPhysicalAddress);
+        }
+
+        /// <summary>
+        /// Get next record
         /// </summary>
         /// <param name="recordInfo"></param>
-        /// <param name="key"></param>
-        /// <param name="value"></param>
-        /// <returns></returns>
-        public bool GetNext(out RecordInfo recordInfo, out Key key, out Value value)
+        /// <returns>True if record found, false if end of scan</returns>
+        public bool GetNext(out RecordInfo recordInfo)
         {
             recordInfo = default(RecordInfo);
-            key = default(Key);
-            value = default(Value);
 
             currentAddress = nextAddress;
             while (true)
@@ -117,7 +132,6 @@ namespace FASTER.core
                     continue;
                 }
 
-
                 if (currentAddress >= hlog.HeadAddress)
                 {
                     // Read record from cached page memory
@@ -129,9 +143,8 @@ namespace FASTER.core
                         continue;
                     }
 
+                    currentPhysicalAddress = _physicalAddress;
                     recordInfo = hlog.GetInfo(_physicalAddress);
-                    key = hlog.GetKey(_physicalAddress);
-                    value = hlog.GetValue(_physicalAddress);
                     nextAddress = currentAddress + hlog.GetRecordSize(_physicalAddress);
                     return true;
                 }
@@ -144,12 +157,33 @@ namespace FASTER.core
                     continue;
                 }
 
+                currentPhysicalAddress = physicalAddress;
                 recordInfo = hlog.GetInfo(physicalAddress);
-                key = hlog.GetKey(physicalAddress);
-                value = hlog.GetValue(physicalAddress);
                 nextAddress = currentAddress + hlog.GetRecordSize(physicalAddress);
                 return true;
             }
+        }
+
+        /// <summary>
+        /// Get next record in iterator
+        /// </summary>
+        /// <param name="recordInfo"></param>
+        /// <param name="key"></param>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public bool GetNext(out RecordInfo recordInfo, out Key key, out Value value)
+        {
+            key = default(Key);
+            value = default(Value);
+
+            if (GetNext(out recordInfo))
+            {
+                key = GetKey();
+                value = GetValue();
+                return true;
+            }
+
+            return false;
         }
 
         /// <summary>
--- a/cs/src/core/Allocator/IFasterScanIterator.cs
+++ b/cs/src/core/Allocator/IFasterScanIterator.cs
@@ -34,6 +34,25 @@ namespace FASTER.core
     public interface IFasterScanIterator<Key, Value> : IDisposable
     {
         /// <summary>
+        /// Gets reference to current key
+        /// </summary>
+        /// <returns></returns>
+        ref Key GetKey();
+
+        /// <summary>
+        /// Gets reference to current value
+        /// </summary>
+        /// <returns></returns>
+        ref Value GetValue();
+
+        /// <summary>
+        /// Get next record
+        /// </summary>
+        /// <param name="recordInfo"></param>
+        /// <returns>True if record found, false if end of scan</returns>
+        bool GetNext(out RecordInfo recordInfo);
+
+        /// <summary>
         /// Get next record
         /// </summary>
         /// <param name="recordInfo"></param>
--- a/cs/src/core/Allocator/VarLenBlittableScanIterator.cs
+++ b/cs/src/core/Allocator/VarLenBlittableScanIterator.cs
@@ -10,21 +10,19 @@ namespace FASTER.core
     /// <summary>
     /// Scan iterator for hybrid log
     /// </summary>
-    public class GenericScanIterator<Key, Value> : IFasterScanIterator<Key, Value>
+    public class VariableLengthBlittableScanIterator<Key, Value> : IFasterScanIterator<Key, Value>
         where Key : new()
         where Value : new()
     {
         private readonly int frameSize;
-        private readonly GenericAllocator<Key, Value> hlog;
+        private readonly VariableLengthBlittableAllocator<Key, Value> hlog;
         private readonly long beginAddress, endAddress;
-        private readonly GenericFrame<Key, Value> frame;
+        private readonly BlittableFrame frame;
         private readonly CountdownEvent[] loaded;
-        private readonly int recordSize;
 
         private bool first = true;
         private long currentAddress, nextAddress;
-        private Key currentKey;
-        private Value currentValue;
+        private long currentPhysicalAddress;
 
         /// <summary>
         /// Current address
@@ -38,7 +36,7 @@ namespace FASTER.core
         /// <param name="beginAddress"></param>
         /// <param name="endAddress"></param>
         /// <param name="scanBufferingMode"></param>
-        public unsafe GenericScanIterator(GenericAllocator<Key, Value> hlog, long beginAddress, long endAddress, ScanBufferingMode scanBufferingMode)
+        public unsafe VariableLengthBlittableScanIterator(VariableLengthBlittableAllocator<Key, Value> hlog, long beginAddress, long endAddress, ScanBufferingMode scanBufferingMode)
         {
             this.hlog = hlog;
 
@@ -47,8 +45,6 @@ namespace FASTER.core
 
             this.beginAddress = beginAddress;
             this.endAddress = endAddress;
-
-            recordSize = hlog.GetRecordSize(0);
             currentAddress = -1;
             nextAddress = beginAddress;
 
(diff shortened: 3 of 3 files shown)


marius-klimantavicius:
Just found that this implementation (based on blittable allocator) does not correctly handle deleted records that are still in memory - when record is deleted in place its value is overwritten by `default`:
```C#
if (entry.word == Interlocked.CompareExchange(ref bucket->bucket_entries[slot], updatedEntry.word, entry.word))
{
  // Apply tombstone bit to the record
  hlog.GetInfo(physicalAddress).Tombstone = true;

  // Write default value
  Value v = default(Value);
  functions.ConcurrentWriter(ref hlog.GetKey(physicalAddress), ref v, ref hlog.GetValue(physicalAddress));

  status = OperationStatus.SUCCESS;
  goto LatchRelease; // Release shared latch (if acquired)
}
```

We effectively lose value and its length so there is no way to find out the true length of a record.

I don't understand how then normal `Read` works correctly but my guess is that index contains valid addresses to records.

marius-klimantavicius:
It seems that there are even more issues with variable length structs (when the records are still in memory)... There is optimization to do an inplace update of records for existing value, however the new value could be longer that previous thus resulting in overflow.

marius-klimantavicius:
Ideas I have right now:
1. Add actual `capacity` record - not needed for fixed size key/value, somehow limit to varlen only.
1. Never overwrite existing records - higher memory usage, might be more issues the with the way memory area is implemented.
1. Append record with _encoded_ padding indicating how much padding is there. This could also be used to align records to 4/8-byte boundaries.

For now I am leaning towards adding padding (only varlen allocator would be affected). Maybe could use scheme similar to PKCS7 padding:
1. 1-byte padding - `01`
1. 2-byte padding - `02 02`
1. 3-byte padding - `03 03 03`
1. 4-byte padding - `04 04 04 04`
1. ... same for paddings up to 7
1. 8-or-more-byte padding - `{FF * n} L3 L2 L1 L0` where `0xL0L1L2L3` is actual padding length. `FF` is repeated up to 4 times so that padding length occurs at 4-byte boundary (aligned to 4 bytes).

marius-klimantavicius:
Updated PR with fix for enumerating deleted records (tombstones that overwrite value) and fix fo #167.

Extended allocator to expose `CanWriteInPlace` to check whether the existing record has enough space.

Added `capacity:int32` (just after `RecordInfo`) to indicate actual record size. Added 8-byte padding to records when using varlen allocator (this might be needed as CLR requires [at least according to spec] natural alignment when accessing built-in types and `RecordInfo` is essentially a long thus should be aligned to 8-bytes).

Note that I haven't checked if RMW is not broken (should not be for fixed length allocators, I am not using it in my case)

badrishc:
Thanks for the PR. We will take a look at this in the coming week. :)

marius-klimantavicius:
Looking into comment in #167 - I think I will add an option to either always create copies or add capacity just after `RecordInfo` (like I do in this PR).

marius-klimantavicius:
Should `AllocatorBase.BeginAddress` be part of checkpoint/recovery?

I was continuing to play around with this (my final goal is to have compaction) and found another issue (not sure if it is with iterator or checkpoints).
What I see is that after compaction the `BeginAddress` is set to some value, but after restart (+recovery) it is set back to initial value of `kFirstValidAddress (64)`.

badrishc:
There will be holes in the log for reasons other than delete as well. For example, if two threads allocate space for the same key, one of them might have to leave the hole and retry.

The way we recommend to skip holes during scan in this case is to move forward one byte at a time and interpret as RecordInfo and check if it's valid. That's what the C++ version, which supports varlen inline, does.

badrishc:
If we avoid adding record size in header, we can also avoid WriteInfo becoming virtual which is more expensive.

badrishc:
Regarding this code in Delete:

```cs
// Write default value
Value v = default(Value);
functions.ConcurrentWriter(ref hlog.GetKey(physicalAddress), ref v, ref hlog.GetValue(physicalAddress));
```

The reason we do the above is so that class objects are deallocated and GCed when a delete occurs. An alternative approach would be to only do this default write for the Generic allocator, since it does not hurt to keep the struct contents there in other cases. Thus we can avoid an additional header field, which is not desirable.

Note that the above-mentioned logic of skipping empty bytes during a scan is still needed, because there may be holes in the log due to failed allocations, as described above.

marius-klimantavicius:
> The way we recommend to skip holes during scan in this case is to move forward one byte at a time and interpret as RecordInfo and check if it's valid. That's what the C++ version, which supports varlen inline, does

This sounds really unsafe or at least unreliable - what if you hit `RecordInfo`-like random bytes (is it guaranteed that hole bytes are `0`?), this will result in random memory read (with possibly garbage data being returned from scan and then skipping over possibly valid records; that's what I was seeing when trying to iterate over deleted records).

> There will be holes in the log for reasons other than delete as well. For example, if two threads allocate space for the same key, one of them might have to leave the hole and retry.

Does allocator fail to write `RecordInfo` in such cases? If it does not write out `RecordInfo` then of course additional field would not help (as it would be missing). But if `RecordInfo` is written out then I'd rather have a setting that specifies whether additional header field is written or not (still working on that). 
 
As for virtual call - there is already one as part of it (`ref GetInfo(physicalAddress)` which was being called to actually get reference to `RecordInfo` to write to) and I'd expect jit to devirtualize it (as it is a virtual call in a `sealed` class) so in the end we still should end up with only one virtual call.

badrishc:
Well, there can be empty spaces in the log at the end of a page, where the next record does not fit. These are zeroed out, and the scanner is expected to skip them by looking for a valid record header. For normal allocations, we write the header, key, and value.

Assume we fix deletes so that only object pointers (generic allocator) are deleted, but inline keys and values are left on the log.

If we let the allocator retain the key and value during deletes, we don't need to waste space on a total record size field. We can simply query the key and value for their lengths, to determine the course of action.

marius-klimantavicius:
Yes, that's was one of the options I was considering. If the records are never overwritten (might allow overwriting if lengths are exactly the same) then there is no need to have `capacity` field (right now this is implemented as a [virtual] call to `CanWriteInPlace`). I have implemented settings to enable/disable `capacity` and alignment in https://github.com/marius-klimantavicius/mister/commit/19d6d5f089f52e9bad4d8f796a4702285d24708b I am not adding it to this PR yet as I might still might remove `capacity` altogether (not sure regarding alignment though) and even though using marker generic arguments is really efficient (at least in .net core) it is somewhat painful to map `bool` to a correct type argument (that's why I have `VariableLengthBlittableAllocator<Key, Value>.Create`).
And I still am unable to find a satisfactory way to deal with holes at the end of page (that does not require guessing where the record starts). Maybe iterator could _know_ that it's at the end of page and could skip to the next page iff all remaining bytes in page are `0`.

badrishc:
We deal with holes in the C++ version as follows: https://github.com/microsoft/FASTER/blob/667434e28f7b1e19498b01486146a836183a2cf9/cc/src/core/faster.h#L1834

For Upserts, handling a size increase using CanWriteInPlace is somewhat similar to what we do in C++, where the PutAtomic() is supposed to return false if the upsert cannot be done in place.

Note that FASTER is multi-threaded with user-level record locking in the mutable region. CanWriteInPlace is opaque to the user. However, the user may want to atomically mark the older record as read-only, which is why we made the user return a bool from PutAtomic() in C++: they can take a record lock, check the size, mark the record as read-only, release the lock, then return false. Unfortunately, this technique in C# would be a breaking change for the signature of functions.ConcurrentWriter.

For Upserts, the atomic marking of read-only is not very important since these are blind upserts and so ordering across threads does not matter. However, handling size increase with RMW is much trickier, since we want to make sure no update is lost (e.g., one thread does read-copy-update, whereas another thread does in place update of old value). This is why we let RmwAtomic() return a bool in C++, so the user can appropriately mark the record as read-only in an atomic manner. For symmetry, PutAtomic() also returns a bool.

marius-klimantavicius:
I will check out C++ version at some point in near future, but at a glance it seems that it expects that there is always at least `sizeof(header)` (8?) bytes in the hole. What if we end up 4-bytes short of the end of page (or in unaligned case 3-bytes short to the end of page) - will look into C++ version, probably will find an answer there.
As for `CanWriteInPlace` I am still trying to understand. It seems that we cannot rely on `KeyLength` and `ValueLength` (unless the user makes sure to correctly handle multiple writers); `CanWriteInPlace` can break RMW if it ever returns true and ideally it would be a decision for user to make (`PutAtomic`/`RmwAtomic`).

marius-klimantavicius:
I kind of have possible breaking implementation (`ConcurrentWriter` and `InPlaceUpdater` returning `bool`). Trying to figure out a way to maintain backwards compatibility (we basically only need that with var len structs). One way I could do that is to check if `Functions` is let's say `IVariableLengthFunctions<...>` and have a condition (not really pleasant to look into):
```C#
if (variableFunctions != null)
{
  if (variableFunctions.ConcurrentWriter(...))
  {
    ...
  }
}
else
{
  functions.ConcurrentWriter(...);
  ...
}
```

I could make breaking changes. Or maybe I can come up with a way to provide a simple wrapper (struct) - the biggest issue with this one is the fact that constructors do not have type inference.

badrishc:
How about this non-breaking approach (similar to what you mention):

(1) Create a type `IVariableLengthFunctions<Key, Value>` with two methods:
* bool CanUpsertInPlace(ref Key key, ref Value src, ref Value dst)
* bool CanRMWInPlace(ref Key key, ref Input input, ref Value value)

(2) Have `IVariableLenthFunctions` be a field in `VariableLengthStructSettings` -- when using varlen, the user has to provide an implementation for this interface, via settings. Thus, the actual `IFunctions` is left untouched. Store it as a readonly variable in FASTER.

(3) Let the main Upsert/RMW code call into the interface if it is non-null If null, assume it is true.

Unrelated, note that with this overall design, the notion of _capacity_, with an additional capacity value in the value header, can be implemented by the user if that makes sense for their payload. It's not hard coded at the FASTER layer.

> Probably should modify VariableLengthAllocator to align to 8 as well.

Exactly.

badrishc:
Second attempt at non-breaking solution. Almost the same as your first proposal, except the new type only implements the new overriding methods.

(1) Create a new type `IVariableLengthFunctions<Key, Value, Input>` that will be used to override the default functions provided in `IFunctions`:
* bool ConcurrentWriter(ref Key key, ref Value src, ref Value dst)
* bool InPlaceUpdater(ref Key key, ref Input input, ref Value value)

(2) Have `IVariableLenthFunctions` be a field in `VariableLengthStructSettings` -- when using varlen, the user has to provide an implementation for this interface, via settings. Thus, the actual `IFunctions` is left untouched. Store it as a readonly variable in FASTER.

(3) Let the main Upsert/RMW code call into the interface if it is non-null If null, fall back to regular `IFunctions`.

User can create a single class that provides both interfaces, if they want.

badrishc:
That said, I am actually leaning towards the breaking solution (change return type to bool) as it is cleaner in the longer term.

It is actually interesting even for the non-varlen case. For example, if a user wants to disable in-place updates for some keys (for instance, they want to log every change to the value), they can simply define the ConcurrentWriter and InPlaceUpdater to simply return false for those keys.

@gunaprsd, any comments on the design, or issues with recovery?

marius-klimantavicius:
Updated PR with partially breaking changes - `IFunctions` remains the same, however `VariableLengthStructSettings` takes in additional type parameter. I can also break `IFunctions` as well.

marius-klimantavicius:
Just pushed a commit with breaking changes to `IFunctions` - I can revert it if you decide to go with `IVariableLengthFunctions`.

badrishc:
It's looking good, thanks! Are you done with the PR or will there be more checkins/testcases?

marius-klimantavicius:
Ideally I would like to add a few unit tests:
1. Skipping null records at the end of page
1. Checking that new records are created if `ConcurrentWriter/InPlaceUpdater` returned `false`

However I won't have time for the next few days, after that I could either modify this PR or just create a new one. 

Otherwise I am more or less satisfied with PR.

marius-klimantavicius:
Added `Delete` method to `IFasterKV`.
Added a few unit tests.
This is now finished from my point of view (unless you need me to change anything). If I find new issues I will create new PR/issue.

marius-klimantavicius:
Just found another issue with iterator - will fix soon.
I this these lines are not correct (both blittable and varlen [which is based on blittable]):
```C#
                var recordSize = hlog.GetRecordSize(hlog.GetPhysicalAddress(currentAddress));
                // Check if record fits on page, if not skip to next page
                if ((currentAddress & hlog.PageSizeMask) + recordSize > hlog.PageSize)
                {
                    currentAddress = (1 + (currentAddress >> hlog.LogPageSizeBits)) << hlog.LogPageSizeBits;
                    continue;
                }
```

We are trying to read record from `hlog` memory regardless whether the address is in memory or on disk.

badrishc:
It was ok for blittable because GetRecordSize is essentially constant, independent of address. But yeah, now that we support varlen, it makes sense to keep them consistent.

PR #177, opened 2019-09-18
Exposing FASTER log as a first-class abstraction

badrishc:
We are exposing our high-speed latch-free log (or write-ahead log) facility as a generally usable first-class citizen, called `FasterLog`. You can append variable-sized chunks (as `byte[]` or `Span<byte>`) to the log. There is no additional header in the log. Users can perform (potentially tailing) pull-based iteration over any range of the log (including until a future log address). Users can truncate from the head of the log. They can also flush the log at very fine granularity, lower than the usual per-page flushing. All aspects of the log (page size, number of pages in memory, segment size on disk, etc.) are tunable. The log works with async code as well, and we will add async versions of the API going forward.

Note that there is no random lookup index on top of this log - if you want such an indexed log, use `FasterKv` as usual.

### Improvements to Epochs

As part of this PR, we have also significantly optimized the epoch framework, removing important performance bottlenecks that now make it feasible to acquire and release epochs at a very fine granularity. We can perform 35 million acquire/release operation pairs per second per thread, with linear scalability. If users are able to pre-acquire a thread for the operations, we are able to achieve 100 million acquire/release operation pairs per second per thread. This improvement now makes it possible to make FASTER work at high performance in task-based async environments.


badrishc:
See example at https://github.com/microsoft/FASTER/blob/19d5d82641965df3b203b2b9e141ba0594051977/cs/playground/FasterLogSample/Program.cs

badrishc:
Note that this log can live on top of any device that implements the `IDevice` abstraction. We provide out of the box device implementations for:
* Generic local storage device (any .NET core system)
* Local storage device (Windows specific)
* Azure page blob storage device
* Tiered storage device (create tiers of storage, log writes to all tiers, read from lowest available tier)
* Sharded storage device (spread writes across multiple devices to get higher effective bandwidth)

badrishc:
Checked in log commit and recovery support. By default, we flush the log at page boundaries. But, the user can manually flush more frequently, e.g., every 10ms. The playground [sample](https://github.com/microsoft/FASTER/blob/fasterlog/cs/playground/FasterLogSample/Program.cs) has been updated to show this facility as well. You can kill and restart the sample as many times as you want, it will auto-resume.

badrishc:
As for end to end performance, for 100 byte payloads, we get around 1GB/sec to a modern NVMe SSD with a single thread, with log tail increment being the bottleneck as we increase the number of append threads. For 1000 byte payloads, we get close to the SSD bandwidth limit (1.8 - 2 GB/sec) using one thread.

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -0,0 +1,321 @@
...
+        /// <param name="entry"></param>
+        /// <returns>Logical address of added entry</returns>
+        public unsafe long Append(Span<byte> entry)
+        {
+            epoch.Resume();
+            var length = entry.Length;
+            BlockAllocate(4 + length, out long logicalAddress);
+            var physicalAddress = allocator.GetPhysicalAddress(logicalAddress);
+            *(int*)physicalAddress = length;
+            fixed (byte* bp = &entry.GetPinnableReference())
```
hiteshmadan: instead of pinning the entry, you can use the CopyTo API (which is also apparently faster in many cases than other memory copy APIs).

entry.CopyTo(new Span<byte>((void*)(4 + physicalAddress), length));

hiteshmadan:
In Iterator.GetNext, there is a code path that allocates a byte[] and copies the read data to it.

1. Does this happen when the read and write are happening on the same shared buffer? If so, it'd be valuable to allow for the allocation to be via a memory pool since this is going to be a pretty frequent occurrence (reader keeping up with the writer).
Please consider accepting an IMemoryPool<byte> and rent an array from that pool. You'll probably have to return an IMemoryOwner<T> instead of a Span<T> though.

2. In code paths where it doesn't allocate, is the span safe to keep around as a reference in the app code? Or does it point to a buffer that (eventually) can be reused for a different page?

cs/src/core/Device/LocalStorageDevice.cs
```diff
@@ -53,14 +57,19 @@ private void RecoverFiles()
...
+            int prevSegmentId = -1;
+            foreach (int segmentId in segids)
+            {
                 if (segmentId != prevSegmentId + 1)
                 {
                     startSegment = segmentId;
-
                 }
                 else
                 {
```
tli2: Do we need to truncate the segments that are not tracked by this device?

cs/src/core/Device/ManagedLocalStorageDevice.cs
```diff
@@ -48,14 +51,19 @@ private void RecoverFiles()
 
             string bareName = fi.Name;
 
-            int prevSegmentId = -1;
+            List<int> segids = new List<int>();
```
tli2: Maybe I am missing something, but it's unclear to me why the list is required. Is it just for readability?

cs/src/core/Index/FasterLog/ILogCommitManager.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// Log commit manager
+    /// </summary>
+    public interface ILogCommitManager
+    {
+        /// <summary>
+        /// Perform (synchronous) commit with specified metadata
+        /// </summary>
+        /// <param name="address">Address committed until (for information only, not necessary to persist)</param>
+        /// <param name="commitMetadata">Commit metadata</param>
+        void Commit(long address, byte[] commitMetadata);
```
tli2: I am wondering if `byte[]` is too low level of an abstraction. Perhaps an interface with a custom serialization/deserialization method is a better choice?

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -0,0 +1,321 @@
...
+            epoch.Suspend();
+            return true;
+        }
+
+        /// <summary>
+        /// Append batch of entries to log
+        /// </summary>
+        /// <param name="entries"></param>
+        /// <returns>Logical address of last added entry</returns>
+        public unsafe long Append(List<byte[]> entries)
```
hiteshmadan: +1 to having a batch append API. However this API forces too many allocations - 1 for the list, and 1 array allocation for each entry in this case:
If the serialized byte array is generated by a serializer that pools its' buffers (like Microsoft Bond), most likely it'll either return the result as an ArraySegment / Span / IMemoryOwner<byte>.

How does something like this look:

public unsafe long Append(int numEntries, Func<int, ArraySegment<byte>> getBytes)

(note that Span<byte> cannot be used here in place of ArraySegment<byte>, details here: https://adamsitnik.com/Span/#span-must-not-be-a-generic-type-argument)

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -0,0 +1,321 @@
...
+            logicalAddress = 0;
+            long tail = -allocator.GetTailAddress();
+            allocator.CheckForAllocateComplete(ref tail);
+            if (tail < 0)
+            {
+                epoch.Suspend();
+                return false;
+            }
+            var length = entry.Length;
+            BlockAllocate(4 + length, out logicalAddress);
```
hiteshmadan: If there is already pending data waiting to be flushed + the incoming entry is big enough to not fit on the space remaining on the current page + no more pages can be allocated, BlockAllocate will block the thread waiting for the allocation to go through.

Is there any way for the early-exit check to account for entry.Length too?

cs/src/core/Index/FasterLog/FasterLog.cs
badrishc: I actually benchmarked the two alternatives and went with pinning as it was non-trivially faster.

cs/src/core/Device/Devices.cs
```diff
@@ -26,9 +26,10 @@ public static class Devices
         /// <param name="logPath">Path to file that will store the log (empty for null device)</param>
         /// <param name="preallocateFile">Whether we try to preallocate the file on creation</param>
         /// <param name="deleteOnClose">Delete files on close</param>
-        /// <param name="capacity"></param>
+        /// <param name="capacity">The maximal number of bytes this storage device can accommondate, or CAPACITY_UNSPECIFIED if there is no such limit</param>
+        /// <param name="recoverDevice">Whether to recover device metadata from existing files</param>
         /// <returns>Device instance</returns>
-        public static IDevice CreateLogDevice(string logPath, bool preallocateFile = true, bool deleteOnClose = false, long capacity = CAPACITY_UNSPECIFIED)
+        public static IDevice CreateLogDevice(string logPath, bool preallocateFile = true, bool deleteOnClose = false, long capacity = CAPACITY_UNSPECIFIED, bool recoverDevice = false)
```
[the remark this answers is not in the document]
badrishc: That makes sense, but the goal was to unblock current users because the recovery path was causing slowdown and incorrectness in certain cases. The goal was to make "no recovery" as the default so current users default to that.

cs/src/core/Device/ManagedLocalStorageDevice.cs
badrishc: The list of files comes in alphabetical order by default, not numerical. Caused a bug without sorting.

cs/src/core/Index/FasterLog/ILogCommitManager.cs
badrishc: I thought of that and even tried to prototype it but it was not as convenient in the end because the user had to implement too many callbacks. This can be revisited though.

badrishc:
> In Iterator.GetNext, there is a code path that allocates a byte[] and copies the read data to it.
> 
> 1. Does this happen when the read and write are happening on the same shared buffer? If so, it'd be valuable to allow for the allocation to be via a memory pool since this is going to be a pretty frequent occurrence (reader keeping up with the writer).
>    Please consider accepting an IMemoryPool and rent an array from that pool. You'll probably have to return an IMemoryOwner instead of a Span though.
> 2. In code paths where it doesn't allocate, is the span safe to keep around as a reference in the app code? Or does it point to a buffer that (eventually) can be reused for a different page?


Correct, when reads and writes share the buffer, we have to copy out because we can only access memory safely under epoch protection. Agreed that memory pool is a possibility. If we had that, a cleaner design might be to always copy out to the user-provided buffer regardless of where we are reading from. This will address your concern in point 2 as well (currently, you must give up the span as soon as you invoke another GetNext).

cs/src/core/Index/FasterLog/FasterLog.cs
hiteshmadan: As an aside - Does the batch partially show up to the iterator while its still being written? Or does the reader only see the first event from the batch after epoch.Suspend() ?

badrishc:
Fyi, there is a check-in coming in to reduce the number of pages needed in memory. This required a couple of internal changes for correctness.

cs/src/core/Index/FasterLog/FasterLog.cs
AlgorithmsAreCool: @hiteshmadan 
I don't understand everything thats going on in this PR, but if you want to use a span instead of an array segment you just need to declare a custom delegate like
```csharp
public delegate ReadOnlySpan<byte> ByteGetter<TState>(int index, TState state);
```
or something like it. That sidesteps the generic parameter constraint.

badrishc: After all the optimizations and inlining, there isn't much benefit to a batched interface in terms of performance, because acquire and release epoch - which were the only calls getting amortized - have been made very inexpensive. Also, we can't allocate for the batch in bulk because the page sizes are pretty small. Maybe we should stick to the unbatched interface?

cs/src/core/Index/FasterLog/FasterLog.cs
badrishc: We only made the common path non blocking right now. Do we need it to always be non blocking? The system atomically allocates space when when it's not yet ready to use, then waits for it to become usable. If we want it to be non blocking, we will need to surface this in the API, i.e. TryAppend will return an incomplete address that the user will need to compete by calling TryCompleteAppend. Is this what you are looking for?

cs/src/core/Device/LocalStorageDevice.cs
badrishc: You mean delete existing files when the device is created? That would depend upon whether the device is being used as new or for recovery.

badrishc:
We now have both low-mem support (as low as 4 pages) as well as truly non-blocking TryAppend.

We also have an async API variant for append (AppendAsync) that returns only when append flush is done, but this is a rough API that needs further feedback/review. There are also significant performance implications of using this interface. Find the prototype here: https://github.com/microsoft/FASTER/pull/180

badrishc:
This PR is almost ready to merge, and definitely ready to try out. Features include:

(1) Support for blocking Append, TryAppend, and AppendAsync
(2) Support for appending byte[] and Span<byte> to log
(3) Log auto-commit at page boundary + user-controlled commit (e.g., every 5ms)
(4) Configurable commit provider (`ILogCommitManager`) that can write commit info anywhere, and hook in pre- and post-commit operations.
(5) Atomic commit of a batch of entries to log (`ISpanBatch`)
(6) Multi-threaded and multi-task appends supported
(7) Tailing iterator over the log
(8) Exposed `CommitTask` so users can await the next commit if they want

See https://github.com/microsoft/FASTER/blob/master/docs/cs/FasterLog.md for a first draft of usage guide, and the comprehensive sample with comments at https://github.com/microsoft/FASTER/blob/fasterlog/cs/playground/FasterLogSample/Program.cs

## Pending

We need to refine/enhance the ways that pooled byte arrays can be provided to the iterator; currently we defined a delegate called `GetMemory` [here](https://github.com/microsoft/FASTER/blob/b06d11253e4bd065b67c8ffbfc04ecaddbf74091/cs/src/core/Index/FasterLog/FasterLogIterator.cs#L16). Any other ideas here?

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -0,0 +1,472 @@
...
+        }
+
+        /// <summary>
+        /// Try to append entry to log. If it returns true, we are
+        /// done. If it returns false, we need to retry.
+        /// </summary>
+        /// <param name="entry">Entry to be appended to log</param>
+        /// <param name="logicalAddress">Logical address of added entry</param>
+        /// <returns>Whether the append succeeded</returns>
+        public unsafe bool TryAppend(Span<byte> entry, out long logicalAddress)
```
AlgorithmsAreCool: Nit: You could make this a `ReadOnlySpan` to communicate semantics a bit better

badrishc: Done.

cs/playground/FasterLogSample/Program.cs
```diff
@@ -87,9 +87,9 @@ static void ScanThread()
             {
                 while (true)
                 {
-                    while (!iter.GetNext(out result))
+                    while (!iter.GetNext(out result, out int length))
                     {
-                        Thread.Sleep(1000);
+                        iter.WaitAsync().GetAwaiter().GetResult();
```
ReubenBond: I wonder if this should be an `IAsyncEnumerable<T>` instead? Can change later, but it is likely a nicer abstraction for log scanning

hiteshmadan: +1! Will that need a .net core 3.0 dependency, though? I see that right now FasterLog targets netstandard20.

ReubenBond: We don't need .NET Core 3.0 for `IAsyncEnumerable<T>`. Orleans uses it and we target netstandard2.0, too. You do need to reference newer libraries, though and you don't get `yield` support AFAIK unless you update to the .NET Core 3.0 *SDK* (possibly bumping to `<LangVersion>8.0</LangVersion>` in the csproj)

badrishc: The problem is that T in our case would be Span<byte>, and we cannot have a Span as a generic parameter.

badrishc: And our iterator returns two things: a span of byte, and an actual length of the item being returned. This allows the user to provide us with a potentially larger span (when we call GetMemory), if they are pooling.

cs/src/core/Index/FasterLog/FasterLogIterator.cs
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Get next record in iterator
+        /// </summary>
+        /// <param name="entry">Copy of entry, if found</param>
+        /// <param name="entryLength">Actual length of entry</param>
+        /// <returns></returns>
+        public unsafe bool GetNext(out Span<byte> entry, out int entryLength)
```
AlgorithmsAreCool: I'm curious if you considered just trimming the `entry` to the precise length instead of returning `entryLength`.
But I'm not sure i fully understand how the buffer return semantics work so maybe that isn't ok

badrishc: The reason I do not trim is that we get the Span from the user via a GetMemory(int minSize) delegate, which may reuse a span that is larger than the requested size. If we returned a truncated span, the caller cannot infer the original size of the span, to be able to reuse in future. Hence, we return the original span and let the user read the first entryLength bytes.

It is a bit convoluted - so am open to other ideas.

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -0,0 +1,472 @@
...
+                logSettings.GetLogSettings(), null, 
+                null, epoch, e => Commit(e));
+            allocator.Initialize();
+            Restore();
+        }
+
+        /// <summary>
+        /// Dispose
+        /// </summary>
+        public void Dispose()
```
hiteshmadan: please do a commitTcs.TrySetException here to release any waiting readers/writers when the log is shutdown?

badrishc: done

badrishc:
## Summary of Interface to FasterLog

```cs
// Enqueue log entry (to memory) with spain-wait

long Enqueue(byte[] entry)
long Enqueue(ReadOnlySpan<byte> entry)
long Enqueue(IReadOnlySpanBatch readOnlySpanBatch)

// Try to enqueue log entry (to memory)

bool TryEnqueue(byte[] entry, out long logicalAddress)
bool TryEnqueue(ReadOnlySpan<byte> entry, out long logicalAddress)
bool TryEnqueue(IReadOnlySpanBatch readOnlySpanBatch, out long logicalAddress)

// Async enqueue log entry (to memory)

async ValueTask<long> EnqueueAsync(byte[] entry)
async ValueTask<long> EnqueueAsync(ReadOnlyMemory<byte> entry)
async ValueTask<long> EnqueueAsync(IReadOnlySpanBatch readOnlySpanBatch)

// Wait for commit

void WaitForCommit(long untilAddress = 0) // spin-wait
async ValueTask WaitForCommitAsync(long untilAddress = 0)

// Commit

void Commit(bool spinWait = false)
async ValueTask CommitAsync()

// Helper: enqueue log entry and spin-wait for commit

long EnqueueAndWaitForCommit(byte[] entry)
long EnqueueAndWaitForCommit(ReadOnlySpan<byte> entry)
long EnqueueAndWaitForCommit(IReadOnlySpanBatch readOnlySpanBatch)

// Helper: enqueue log entry and async wait for commit

async ValueTask<long> EnqueueAndWaitForCommitAsync(byte[] entry)
async ValueTask<long> EnqueueAndWaitForCommitAsync(ReadOnlyMemory<byte> entry)
async ValueTask<long> EnqueueAndWaitForCommitAsync(IReadOnlySpanBatch readOnlySpanBatch)

// Truncate log (from head)

void TruncateUntil(long untilAddress)

// Scan interface

FasterLogScanIterator Scan(long beginAddress, long endAddress)

// FasterLogScanIterator interface

bool GetNext(out byte[] entry, out int entryLength)
bool GetNext(MemoryPool<byte> pool, out IMemoryOwner<byte> entry, out int entryLength)
async ValueTask WaitAsync()

// Random read

async ValueTask<(byte[], int)> ReadAsync(long address, int estimatedLength = 0)
```

badrishc:
Added checksum support for log verification during scan/read. Enable by setting FasterLogSettings.LogChecksum.

badrishc:
Update: we are in the final stages of this PR. Working on correctly handling and surfacing exceptional cases, such as transient and permanent storage failures. This is WIP in a branch (fasterlog-exceptions). ETA within a couple of days.

badrishc:
IAsyncEnumerable support has now been added for the iterator.

badrishc:
In addition to the previously supported TruncateUntil on the log, we now support persistent iterators. You can create any number of iterators over the log, and "name" them if you need them to be part of commits. During recovery, if you create an iterator with the same name, we will resume iteration from the last committed iterator location. Example:

```cs
using (iter = log.Scan(log.BeginAddress, long.MaxValue, name: "foo"))
   await foreach ((byte[] result, int length) in iter.GetAsyncEnumerable())
   {
      ...
   }
```

badrishc:
Detailed documentation is now available at:
* https://github.com/microsoft/FASTER/blob/master/docs/cs/FasterLog.md
* https://microsoft.github.io/FASTER/docs/fasterlog

badrishc:
Merging to master and closing PR as the functionality is complete at this point. We can continue discussions here, and create new PRs for further enhancements as well.

PR #180, opened 2019-09-30
Fasterlog async

badrishc:
Prototype task async version of FasterLog.

### DIFF
--- a/cs/src/core/Index/FasterLog/FasterLog.cs
+++ b/cs/src/core/Index/FasterLog/FasterLog.cs
@@ -4,9 +4,6 @@
 #pragma warning disable 0162
 
 using System;
-using System.Collections.Concurrent;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Runtime.CompilerServices;
 using System.Threading;
@@ -103,10 +100,8 @@ namespace FASTER.core
         }
 
         /// <summary>
-        /// Try to append entry to log. If is returns true, we are
-        /// done. If it returns false with negative address, user
-        /// needs to call TryCompleteAppend to finalize the append.
-        /// See TryCompleteAppend for more info.
+        /// Try to append entry to log. If it returns true, we are
+        /// done. If it returns false, we need to retry.
         /// </summary>
         /// <param name="entry">Entry to be appended to log</param>
         /// <param name="logicalAddress">Logical address of added entry</param>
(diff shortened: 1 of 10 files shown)


cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -179,6 +180,40 @@ public unsafe bool TryAppend(Span<byte> entry, ref long logicalAddress)
             return true;
         }
 
+        private TaskCompletionSource<long> commitTask = new TaskCompletionSource<long>();
```
hiteshmadan: Please use TaskCreationOptions.RunContinuationsAsynchronously to avoid potential deadlocks in consuming code?

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -179,6 +180,40 @@ public unsafe bool TryAppend(Span<byte> entry, ref long logicalAddress)
...
+        public async ValueTask<long> AppendAsync(byte[] entry)
+        {
+            long logicalAddress = 0;
+
+            while (true)
+            {
+                if (TryAppend(entry, ref logicalAddress))
+                    break;
+
+                await commitTask.Task;
```
hiteshmadan: since appends are concurrent, any need to protect this commitTask access with a lock / interlocked.CompareExchange ? We don't want to accidentally use a stale commitTask that's already in completed state due to a past commit ?

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -368,6 +425,10 @@ private void Commit(long flushAddress)
                     CommittedUntilAddress = flushAddress;
                     // info.DebugPrint();
                 }
+
+                var _commitTask = commitTask;
+                commitTask = new TaskCompletionSource<long>();
+                _commitTask.SetResult(flushAddress);
```
hiteshmadan: is this SetResult guaranteed to happen only once since commit's are done on a single background thread?
If not, a second (accidental) call to SetResult will result in an InvalidOperationException.

badrishc: Since SetResult happens under a lock (per commit), we should be fine.

cs/src/core/Index/FasterLog/FasterLog.cs
badrishc: We order the memory access so that the task is cached, then the commit address is read. Since the committer first updates commit address, then updates the task, we should be fine. Using a stale task is fine because all state tasks will have a completed status, which will cause this loop to retry.

PR #183, opened 2019-10-10
Fix nuget links

Tornhoof:
the package is now hosted at https://www.nuget.org/packages/Microsoft.FASTER/


badrishc:
Yes, thanks for fixing the link!

PR #184, opened 2019-10-17
Fix: handle CPR_SHIFT_DETECTED and RETRY_LATER statuses in Delete operation.

1u0:
Hi!
If I understand the code correctly, the C# version of FASTER doesn't handle `CPR_SHIFT_DETECTED` and `RETRY_LATER` `OperationStatus`es in `Delete` method.


badrishc:
Closing as the issue is being handled separately (the master has diverged significantly).

PR #188, opened 2019-10-26
cc: GetAtomic: change return type to bool

matthewbrookes:
I'm a bit unsure on this but I'm wondering whether it's possible for GetAtomic to read a stale value if another thread has concurrently "replaced" the record? We mark variable-length values as replaced when they are made to be larger than the available buffer in the value. I think the following trace is possible:

* Threads R (reader) and U (upserter) look up the value at address X
* U acquires a user-level lock on value
* U tries to perform upsert but new value is too large for buffer
* U marks record as replaced and releases lock on X 
* U creates a new record at address Y
* R reads the value at X (which is stale)

By changing the return type to `bool`, R can detect (via a flag in Value) that record has been replaced and can retry the operation. R then reads the newer record.

What do you think @badrishc?

### DIFF
--- a/cc/benchmark-dir/benchmark.cc
+++ b/cc/benchmark-dir/benchmark.cc
@@ -141,7 +141,7 @@ class ReadContext : public IAsyncContext {
 
   // For this benchmark, we don't copy out, so these are no-ops.
   inline void Get(const value_t& value) { }
-  inline void GetAtomic(const value_t& value) { }
+  inline bool GetAtomic(const value_t& value) { return true; }
 
  protected:
   /// The explicit interface requires a DeepCopy_Internal() implementation.
--- a/cc/playground/sum_store-dir/sum_store.h
+++ b/cc/playground/sum_store-dir/sum_store.h
@@ -142,8 +142,9 @@ class ReadContext : public IAsyncContext {
   inline void Get(const value_t& value) {
     *result_ = value.num_clicks;
   }
-  inline void GetAtomic(const value_t& value) {
+  inline bool GetAtomic(const value_t& value) {
     *result_ = value.atomic_num_clicks;
+    return true;
   }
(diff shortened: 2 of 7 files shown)


1u0:
Just my 2c, from my understanding the `Upsert` contract in C++ version follows roughly the steps:
* U (upserter) look up the value at address X;
* U _may read_ the old value, but only to determine the old buffer size;
* if the old buffer is too small, return `false` in `PutAtomic()` and fallback to new entry creation using `Put()` (X is not locked);
* otherwise, place a new value at X atomically, so that other operations see it consistently.

In general, if _`V`_ is the initial value and _`VU`_ is the value after an upsert, it's valid behavior if the reader R returns either _`V`_ or _`VU`_ in this case.

However, maybe there is a valid case for `bool GetAtomic()` in some situations.

matthewbrookes:
Yeah @1u0 that's certainly the behaviour now. What would be nice would be for the reader to be able to detect that the value is stale and then retry to get the newer value.

What is somewhat annoying is that this would be a breaking change and for most users it will be sufficient to just `return true` from the method.

If it's decided to go along with this change then some explanatory comments etc can be added.

badrishc:
This is an interesting variant. That said, I agree with the points made by @1u0. Apart from breaking current semantics, there are two reasons:

* There isn't anything _incorrect_ about the semantics as they stand now. The read is still correct/consistent with respect to FASTER's semantics.
* Users can actually implement the semantics you suggest on top of what we have right now. In other words, the writer can set a field (e.g., in the lock) that indicates to readers that the value is now stale. Then the reader can simply retry the operation to get the latest value.

matthewbrookes:
These are good points :) I was a bit uncertain around the semantics so I'm happy to have clarified these.

PR #193, opened 2019-11-03
Union with "control" variable approach is generalized into value_cont…

wtom76:
…rol and value_atomic_control base classes.

All classes under "cc" directory excluding "test", that are designed with this approach, are refactored to be derivatives of value_control/value_atomic_control.

An attempt to get rid from level 4 warnings is made excluding tests.


badrishc:
Thanks for the contribution. Looks like there are conflicts with the base branch. Could you update accordingly based on latest merge from master.

wtom76:
> Thanks for the contribution. Looks like there are conflicts with the base branch. Could you update accordingly based on latest merge from master.

Hi badrishc,
Sure. Done.
Please run all cc tests multiple times. I've noticed that _assert(value().reserved_ == 0);_ in _Address_ ctor in _address.h_ sometimes failed in both original and mine versions, but can't figure out why.
Also I upgraded VC to 2019 and now build fails due to "C1189: #error:  The <experimental/filesystem> header providing std::experimental::filesystem is deprecated by Microsoft and will be REMOVED."

1u0:
Hi @wtom76,

> Please run all cc tests multiple times. I've noticed that assert(value().reserved == 0);_ in Address ctor in address.h sometimes failed in both original and mine versions

Can you give more details which test (or some example program) fails with this assertion error? And how frequently this occurs?

wtom76:
Hi @1u0,
> Can you give more details which test (or some example program) fails with this assertion error? And how frequently this occurs?

Approximately 1 time per 10 runs. Can't remember right now which test, will look into it at home.

wtom76:
Hi @1u0,
Have run tests yesterday, the issue is not reproduced. Will try one more time. May be it makes sense to run tests in debugger, but I'm not able to do that with VC 2019 at the moment though.

1u0:
I've stumbled this assert failure too by a chance on [CI run](https://dev.azure.com/ms/FASTER/_build/results?buildId=49166):
```
[ RUN      ] PagingTest_ThreadPool.Rmw_Concurrent
Assertion failed: reserved_ == 0, file d:\a\1\s\cc\src\core\address.h, line 57
```
but don't know the root cause yet.

1u0:
Regarding the PR itself:
I think, instead of the proposed change, it would be more simpler to
 * get rid of `AtomicSomeStructure` like classes (use `std::atomic<SomeStructure>` explicitly). This makes the proposed `value_atomic_control` not needed;
 * try to use more type safe api, eliminate or minimise usage of "union style" casts (instead of building helper code around it).

Even if it happens that "union style" casts are still would be present, I think there is an alternative way to provide a generic helper function that can do "union style" cast. In that case `SomeStructure` classes would continue to be domain types that also define binary layout.

As cons, the change proposed here is intrusive (forces to inherit from `value_control`) and also spreads domain types into too many (redundant, imo) classes.

badrishc:
I agree that forcing to inherit from `value_control` is rather intrusive and I felt it to be less understandable than before.

badrishc:
Closing PR for now - we can revisit in future if this makes sense for customers.

PR #198, opened 2019-11-09
[FasterLog] Concurrent iteration

badrishc:
This PR adds support for concurrent iteration. We invoke Scan on FasterLog, and the resulting iterator may be shared by multiple threads iterating concurrently.

We also add an explicit notion of `CompletedUntilAddress` for a persisted iterators, and introduce a gesture `CompleteUntil` on the iterator in order to signify the completion of iteration work until the specified address in the scan. To persist iterator until a given address, use the following pattern:

```cs
iter.GetNext(...); // iterate
// do work on retrieved entry
iter.CompleteUntil(iter.NextAddress); // set completed-until address
await log.CommitAsync(); // commit to ensure persistence of completed-until
```

With concurrent iteration, the consumer is responsible for invoking `CompleteUntil` with an address such that all addresses before it have been completed by the concurrent threads.


badrishc:
PR is ready to review and merge.

We get higher scan performance than before, on one thread, with this implementation. We are able to hit 2GB/sec (close to SSD read bandwidth) with scan on one thread in the FasterLogSample project. The bottleneck is the code to check the entry byte array correctness. We bottleneck SSD bulk read bandwidth (around 2.5GB/sec) on adding the second thread.

Further, when user work on scan results is expensive, we get close to linear scalability as we increase the number of iterator threads (on the same scan iterator).

PR #201, opened 2019-11-12
add a unit test to verify data recovery on close/reopen in windows/linux

hiteshmadan:
I'm seeing a failure in my linux unit tests where i do create log->write 1 entry->close->open->read. Never fails on windows. Also, doesn't repro if i replace write 1 entry with write 2 entries!
Thought i'll add a unit test here since recovery wasn't being covered yet. Will also help direct my bug chase.


badrishc:
Closing for now - pls reopen if the issue persists in latest.

PR #215, opened 2019-12-03
Cloud Checkpoint and Log Commit Manager 

tli2:
Implemented a checkpoint and log commit using cloud storage, completing the device package.

Also adds some refactoring for the original code.

More testing is needed before merge. Opening this PR now so it is visible and I don't forget.


badrishc:
The better way to architect this is to have a checkpoint and commit manager implementation that can work with a general `IDevice` as input. Then we simply instantiate it with AzurePageBlobDevice. That way, the code will be usable with any `IDevice` we add for other storage in the future.

tli2:
I tried that out and I don't think it's as straightforward. The current `IDevice` interface encapsulates no information about how to recover one (nor should it). `ICheckpointManager` currently deals with recovery on startup because it has higher level information (e.g. directories for local device, containers for cloud device, plus naming conventions) to create `IDevice` from on recovery.

If we want to make an `ICheckpointManager` that takes in `IDevice` we will have to push that responsibility up into user code. Either we ask users to create the device on startup, or we ask for some sort of 'DeviceFactory' abstraction that encapsulates the directory/container information. I don't think either is more elegant.

badrishc:
I was thinking more along these lines:

* Create class DeviceCheckpointManager : ICheckpointManager
* This takes as parameter a Func, roughly (string path, string name) => IDevice
* DeviceCheckpointManager handles the path and name creation for various files
* We provide out-of-the-box minimal funcs for local storage device and azure storage device - these are much simpler than entirely new ICheckpointManager instances.
* For GetLatestCheckpoint, we need user to provide a Func that returns an iterator of valid subfolder paths in timestamp order, so we can find the latest valid checkpoint?

tli2:
I have written a `NamedDeviceFactory` abstraction for this, and it works well for the `ILogCommitManager`.

I am a little confused about the complexity of `ICheckpointManager`. The separation of initialization and commit seems to very implementation specific. There is overlapping functionality between `ICheckpointManager` and `ILogCommitManager`, that I assume is there because `ICheckpointManager` was written before the `FasterLog` abstraction. 

Should I try to trim `ICheckpointManager` down a little bit?

badrishc:
What simplification do you have in mind? We can discuss the proposal here to see if it's a fundamental point. Need to also be cognizant of not breaking existing code.

tli2:
For start I was thinking about removing `InitializeIndexCheckpoint` and `InitializeLogCheckpoint`. This seems overly specific to the fact that local checkpoints create folders. I didn't have to do that for cloud ones. 

Is there a valid execution trace where we need to delete all pervious files in a checkpoint folder, or is that more of a precaution against maybe reusing of folders across execution runs?

Going forward maybe the right thing to do is to have a `IndexCheckpointManager` and `LogCheckpointManager`, and only have the functionality of `GetLatestCheckpoint` in the overarching one. Right now having overlapping functionality replicated in both `CheckpointManager` and `LogCommitManager` feels strange. (as I said I suspect that's because `LogCommitManager` was added later)

tli2:
Stale. Closing for now.

Issue #216, opened 2019-12-04
[C#] Variable Length Classes Cant Be Recovered After Process Restart


ThiagoT1:
Hello, FASTER!

I'm trying to use FasterKV and recover data after a checkpoint is taken and the process is restarted.

However, sometimes the checkpoint will get "corrupted" and no data can be recovered, while my deserializers explode.

I put together a minimal repro.

The example below works when writing 20 items, while failing when writing 21.

The error with 21 items is this:

![image](https://user-images.githubusercontent.com/16569217/70105734-766f6d80-1620-11ea-9858-6002b329700a.png)

```csharp

using FASTER.core;
using Microsoft.VisualStudio.TestTools.UnitTesting;
using System;
using System.Collections.Generic;
using System.IO;
using System.Text;

namespace Tests.FasterClassSample
{

    [TestClass]
    public class ClassSample
    {
        static int iterations = 20; // changing this to 21 will make the test fail

        string FasterFolderPath
        {
            get
            {
                return Path.Combine(Path.GetTempPath(), "Faster");
            }
        }

        [TestInitialize]
        public void Cleanup()
        {
            Directory.Delete(FasterFolderPath, true);
        }


        [TestMethod]
        public void ClassSampleTest()
        {
            string logPath, objPath;
            IDevice log, objlog;
            FasterKV<MyKey, MyValue, MyInput, MyOutput, MyContext, MyFunctions> h;
            MyContext context;

            Prepare(out logPath, out objPath, out log, out objlog, out h, out context);

            h.StartSession();

            Write(h, context);

            h.Refresh();

            Read(h, context, false);

            h.TakeFullCheckpoint(out Guid CheckPointID);
            h.CompleteCheckpoint(true);

            Destroy(log, objlog, h);

            Prepare(out logPath, out objPath, out log, out objlog, out h, out context);

            h.Recover();

            h.StartSession();

            Read(h, context, true);

            Destroy(log, objlog, h);


        }

        private void Prepare(out string logPath, out string objPath, out IDevice log, out IDevice objlog, out FasterKV<MyKey, MyValue, MyInput, MyOutput, MyContext, MyFunctions> h, out MyContext context)
        {
            logPath = Path.Combine(FasterFolderPath, $"FasterRecoverTests.log");
            objPath = Path.Combine(FasterFolderPath, $"FasterRecoverTests_HEAP.log");
            log = Devices.CreateLogDevice(logPath);
            objlog = Devices.CreateLogDevice(objPath);
[...]

PR #218, opened 2019-12-04
Fix to snapshot checkpoint with object log, where snapshot objects en…

badrishc:
…d up overwriting each other.

Fix https://github.com/microsoft/FASTER/issues/216

### DIFF
--- a/cs/src/core/Allocator/AllocatorBase.cs
+++ b/cs/src/core/Allocator/AllocatorBase.cs
@@ -352,7 +352,9 @@ namespace FASTER.core
         /// <param name="result"></param>
         /// <param name="device"></param>
         /// <param name="objectLogDevice"></param>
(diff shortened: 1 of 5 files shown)


cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -1366,6 +1368,7 @@ public void AsyncFlushPages(long fromAddress, long untilAddress)
         public void AsyncFlushPagesToDevice(long startPage, long endPage, long endLogicalAddress, IDevice device, IDevice objectLogDevice, out CountdownEvent completed)
         {
             int totalNumPages = (int)(endPage - startPage);
+            var localSegmentOffsets = new long[SegmentBufferSize];
```
marius-klimantavicius: Wondering if we could prevent allocation for non-GenericAllocator case.

badrishc: Hmm, good point, yes, that's doable with some refactoring

Issue #232, opened 2020-01-12
After several checkpoints recovery fails with big data


dangershony:
Still having issues with my POC.

In the code example I push a lot of data to the db and also delete some of the data, then take checkpoints regularly, then recover and continue.

I have the POC here (I can also submit it as a PR to the playground area)
https://github.com/dangershony/FASTER/tree/faster-test/cs/playground/ClassRecoveryDurablity

Just hit run and observe it fail after a few loops.
Note: comment out the method `TestData`  code and call `dotnet run 100` will for some reason work fine but `dotnet run 10` will fail after a few calls to recover.

A quick summery of why I want to use FASTER, I contribute at blockcore, blockchain project, that implements among other things the bitcoin blockchain.
bitcoin uses leveldb in C++ as it's k/v store to persist chain data, I would like to use FASTER for that purpose, the POC will try to simulate the same behavior as a bitcoin node.

badrishc:
Thanks for the detailed info and repro. The issue was that our recent change to support client sessions resulted in skipping a phase (WAIT_PENDING) during checkpointing. This phase was writing out some important metadata (object log offsets), which resulted in us overwriting existing objects during recovery and trying to de-serialize spuriously.

The fix is now pushed to master in the FASTER repo. Please try it out and let us know.

After you are done with the POC, it might be interesting to contribute PR as a playground sample as well.

dangershony:
ah brilliant, sure I will give it a try and when done I can submit the PR to the playground.

dangershony:
Ok this seems to have worked, I will try a few more variants of the my requirements and if all is well I will submit a PR to the playground referencing this issue.

Thanks.

dangershony:
FYI stating the obvious and t's probably time available dependent but its probably a good idea to have test coverage for the specifics that caused this issue.

dangershony:
Another question, I guess its safe to delete old checkpoints I have no need for.
Basically to recover I just need to have the most recent checkpoint right?

badrishc:
Sounds good, yes we will add this to the test coverage.

Correct, to recover you need some checkpoint (combination of log + compatible earlier or same-time index checkpoint). Other checkpoints can be deleted. You also of course need the main log itself.

badrishc:
Added to test coverage now.

PR #233, opened 2020-01-13
Faster test

dangershony:
Adding this project to the play gourd.

The project adds a lot of data to store and also does deletes. 
Test data is generated deterministicly.
Checkpoints and recovery are called very often to  
Indexes are byte arrays 
Test data values are different sizes

https://github.com/microsoft/FASTER/issues/232


dangershony:
Ah yes brilliant thanks for fixing the PR.

PR #237, opened 2020-01-31
inline gtest print support function defns

wilsonqin:
We have some application code including faster.h, and found that inline definitions for the new gtest print support functions reduce our build situations of linker error for multiple defns. There is a decent chance for collisions on the "<<" operator defn depending on what other c++ libraries are included in the application code.

Linux, C++, g++ 7.4

### DIFF
--- a/cc/src/core/faster.h
+++ b/cc/src/core/faster.h
@@ -2869,15 +2869,15 @@ bool FasterKv<K, V, D>::GrowIndex(GrowState::callback_t caller_callback) {
 }
 
 // Some printing support for gtest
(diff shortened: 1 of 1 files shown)


badrishc:
Thanks, Wilson. @Skyprophet, any comments?

Skyprophet:
It is reasonable, I would merge it.

P.s. Modern compilers are usually smart enough to inline them with higher optimization.

wilsonqin:
@Skyprophet, thank you for the feedback. 
I thought compiler optimization would resolve too. Was using -O2 on the application. 
In some builds we have a multiple header include - perhaps that could complicate the optimization.

Issue #238, opened 2020-02-01
[C#] Could Disk Reads be FASTER?


ThiagoT1:
I've got pretty much every basic operation covered on my POC.

Repo: https://github.com/ThiagoT1/FasterDictionary/tree/unsafe-0.0.3-fastread

Iteration thru the whole KV has ok(ish) performance.

```csharp
public IAsyncEnumerator<KeyValuePair<TKey, TValue>> GetAsyncEnumerator(CancellationToken cancellationToken = default)
{
	return new AsyncEnumerator(this);
}
class AsyncEnumerator : IAsyncEnumerator<KeyValuePair<TKey, TValue>>
{
	FasterDictionary<TKey, TValue> _fasterDictionary;

	public ValueTask StartTask { get; }

	public AsyncEnumerator(FasterDictionary<TKey, TValue> fasterDictionary)
	{
		_fasterDictionary = fasterDictionary;
		StartTask = _fasterDictionary.AquireIterator();
	}

	KeyValuePair<TKey, TValue> _current;
	public KeyValuePair<TKey, TValue> Current => _current;

	public ValueTask DisposeAsync()
	{
		return _fasterDictionary.ReleaseIterator();
	}

	public async ValueTask<bool> MoveNextAsync()
	{
		await StartTask;

		_current = default;
		var readResult = await _fasterDictionary.IteratePair();
		if (!readResult.Found)
			return false;

		_current = new KeyValuePair<TKey, TValue>(readResult.Key, readResult.Value);
		return true;
	}
}
```
https://github.com/ThiagoT1/FasterDictionary/blob/96407cf3e476f612eb2eb72e895af4c6f9358850/src/FasterDictionary/FasterDictionary.Iterator.cs#L68-L71

# Issue 

What's killing me now is the random key read performance. I'm talking "mono-threaded" access to KV where a single reader has to retrieve values for thousands of random keys.

Adding, Updating once and Iterating over 1M keys takes less than 2 minutes:

```csharp
[InlineData(1_000_000)]
public async Task AddUpdateIterate(int loops)
{
	FasterDictionary<string[], string>.ReadResult result;
	using (var dictionary = new FasterDictionary<string[], string>(TestHelper.GetKeyComparer<string[]>(), GetOptions($"{nameof(AddIterate)}-{loops}")))
	{
		var keys = new List<string[]>();

		for (var i = 0; i < loops; i++)
		{
			keys.Add(new[] { i.ToString() });
[...]

PR #240, opened 2020-02-04
User should be able to provide context class when using async methods

### LINKED ISSUE #239: [C#] ClientSession Async Methods Fail Without Context
If you try to use any ClientSession<,,,,> async method (Read, RMW, Updsert, Delete) and you Functions class rely on the provided Context class for any callback, you'll get a Null Reference Exception.

ThiagoT1:
Attempts to fix #239 .

If the Functions class depends on the Context instance, async methods will make it fail with a Null Reference Exception.

As per #238, full adherence to TLP semantics would see callbacks on the Functions class gone, maybe replaced by injected middleware semantics, ASPNET.CORE style.

### DIFF
--- a/cs/playground/FasterKVAsyncSample/Program.cs
+++ b/cs/playground/FasterKVAsyncSample/Program.cs
@@ -40,7 +40,7 @@ namespace FasterKVAsyncSample
                 Console.WriteLine($"Unobserved task exception: {e.Exception}");
                 e.SetObserved();
             };
-            
+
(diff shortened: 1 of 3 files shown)


badrishc:
The reason we did not include Context in the async API is that there seemed to be no good use case for contexts in the async world, since the call stack is available after the await.

But yes, I agree it still makes sense to let users provide it because `Functions` expect the same.

ThiagoT1:
We could merge this so users can switch seamlessly between blocking and async code, then.

PR #241, opened 2020-02-04
Thread.Yield on CompletePending control flow is too CPU Heavy

ThiagoT1:
As per #238, frequent calls to CompletePending get a lot lighter on the CPU if we don't Yield the Thread.

Also, there's other call sites to Thread.Yield, that should be justified:

https://github.com/microsoft/FASTER/blob/8f9cf2c2ea8ea334b94ee1585f3cf4cc4cf496ca/cs/src/core/Allocator/ErrorList.cs#L39

https://github.com/microsoft/FASTER/blob/8f9cf2c2ea8ea334b94ee1585f3cf4cc4cf496ca/cs/src/core/Index/FASTER/FASTERImpl.cs#L1883-L1901

https://github.com/microsoft/FASTER/blob/8f9cf2c2ea8ea334b94ee1585f3cf4cc4cf496ca/cs/src/core/Allocator/AllocatorBase.cs#L1430

https://github.com/microsoft/FASTER/blob/8f9cf2c2ea8ea334b94ee1585f3cf4cc4cf496ca/cs/src/core/Index/FasterLog/FasterLog.cs#L948

https://github.com/microsoft/FASTER/blob/8f9cf2c2ea8ea334b94ee1585f3cf4cc4cf496ca/cs/src/core/Index/FasterLog/FasterLog.cs#L1028

### DIFF
--- a/cs/src/core/Index/FASTER/FASTERThread.cs
+++ b/cs/src/core/Index/FASTER/FASTERThread.cs
@@ -147,6 +147,7 @@ namespace FASTER.core
 
         internal bool InternalCompletePending(FasterExecutionContext ctx, bool wait = false)
         {
+            var spin = new SpinWait();
             do
             {
                 bool done = true;
@@ -181,7 +182,7 @@ namespace FASTER.core
                 if (wait)
                 {
                     // SpinOnce before checking again
-                    new SpinWait().SpinOnce();
+                    spin.SpinOnce();
                 }
             } while (wait);
 
--- a/cs/src/core/Index/FASTER/FASTERThread.cs
+++ b/cs/src/core/Index/FASTER/FASTERThread.cs
@@ -180,8 +180,8 @@ namespace FASTER.core
 
                 if (wait)
                 {
-                    // Yield before checking again
-                    Thread.Yield();
+                    // SpinOnce before checking again
+                    new SpinWait().SpinOnce();
                 }
             } while (wait);
 


badrishc:
We did have Spin originally, but we converted it to thread yield as spin was causing higher latencies, I think. We were also unable to saturate NVMe ssd with thread yield. Adding @peterfreiling and @hiteshmadan who may remember the details better.

cs/src/core/Index/FASTER/FASTERThread.cs
```diff
@@ -180,8 +180,8 @@ internal bool InternalCompletePending(FasterExecutionContext ctx, bool wait = fa
 
                 if (wait)
                 {
-                    // Yield before checking again
-                    Thread.Yield();
+                    // SpinOnce before checking again
+                    new SpinWait().SpinOnce();
```
marius-klimantavicius: I think SpinWait is intended to be reused outside the loop as it has state (count) which is used to decided between regular spin and yield.

ThiagoT1: I moved it. Thank you.

ThiagoT1:
> 
> 
> We did have Spin originally, but we converted it to thread yield as spin was causing higher latencies, I think. We were also unable to saturate NVMe ssd with thread yield. Adding @peterfreiling and @hiteshmadan who may remember the details better.

SpinOnce will eventually Yield the thread. Also, less CPU load from this task would benefit other work waiting to be done overall.

peterfreiling:
@ThiagoT1 @badrishc , yes, we replaced the SpinWait calls in CompletePending with Yield, as it greatly improved performance and throughput. IIRC the theory was that the SpinWait was throttling the I/O before the disk was saturated. I originally found this and made the change for the FastBI workflow, which is no longer a production workflow. However, Badrish verified the performance improvement with a perf test (here's a quote from 4/4/2019):

> [4/4/19 5:34 PM] Badrish Chandramouli
>     replacing Thread.SpinWait(100) with Thread.Sleep(1) gives excellent results... Yield() is equally good... removing loop gives 55Kops/sec, Thread.Yield or Sleep with loop gives 77K ops/sec... SpinWait with loop gives 10K ops/sec... I have a testbed to vary this now

So, these gains were not trivial. @Badrishc, do you still have the testbed you were using for this? I would be very hesitant to change this back without fully understanding the implications.

ThiagoT1:
If that was before async methods, it could be the case the Yield was making it possible to schedule disk I/O in a parallel fashion.

The same should be possible with async/TPL, without the cost of so many context switches.

I would suggest a test with full non blocking async code (Yeild's or Spins wouldn't even apply) to

peterfreiling:
Yes this was from April 2019 so well before the async overhaul. Thinking back I believe we were also focused on a SpinWait call in a loop within a different API (when we were initiating a disk read, not complete pending) and then replaced some other SpinWait calls with Yield as well. If we can verify this doesn't regress the original perf gains I'm all for it. Badrish should hopefully still have the test bed.

badrishc:
Yes, the change mentioned above was pre-async. There were non-trivial performance benefits of moving to Thread.Yield when using the sync API with thread-affinitized sessions. There are still users who prefer this vs. async, so we would need to keep that in mind.

> I would suggest a test with full non blocking async code (Yeild's or Spins wouldn't even apply) to rule that out.

Can you clarify what you suggest here?

ThiagoT1:
Do the test that was done before (with a decent amount of input parallelism), but replacing Read with await ReadAsync where possible, and staying away from all blocking (synchronous) API.

Currently the Thread.Yield happens when the (first) iteration inside ```FaskterKV<,,,,,>.InternalCompletePending``` does not set ```done``` to true. Then the Thread is costly sent away to do other work.

In the async methods, this will hit a similar situation when any await does not short circuit and the AsyncStateMachine kicks in.

The difference is that in the async API version the Current Thread will be assigned other work without the overhead of a context switch, albeit with some GC burden from the when the short circuit does not happen.

The point is that the async/await rabbit hole could be FASTER :suspect: than the Context Switch induced by the Thread.Yeild() rabbit hole.

BTW, how is that parallelized load generated? It should also be started using Tasks, avoiding direct spawning of Threads, so the ThreadPool handles everything, as it would probably do in most real usage scenarios.

PR #242, opened 2020-02-04
Ref keys on async methods

ThiagoT1:
Async methods do not support ref Keys, wich are needed when using the unsafe var length strategy.

They could.


badrishc:
Ah, nice workaround. I wonder if we can have a quick path for a synchronous operation that avoids calling the helper.

badrishc:
https://github.com/microsoft/FASTER/blob/8f9cf2c2ea8ea334b94ee1585f3cf4cc4cf496ca/cs/src/core/Index/FasterLog/FasterLog.cs#L237

ThiagoT1:
The same style of nested slow methods was used and its rebased onto master.

PR #248, opened 2020-02-21
Issue 246 continuations wiring (II)

### LINKED ISSUE #246: [C#] Continuations Should be Properly Wired
Currently, ```ReadAsync``` continuations will try to complete  every pending read.
This leads to unnecessary race conditions  on the ```FasterExecutionContext```members.

https://github.com/microsoft/FASTER/blob/bd5ad88df06e881559f56fd2b3c085e12f3229bf/cs/src/core/Index/FASTER/FASTERThread.cs#L258-L262

I sense this is like this since KVSession is not (should it be?) thread safe, maybe still in a gray area between the older thread affinity from FasterKV and the WIP adoption of TPL.
However, when using the async methods, there's nothing preventing multi-threaded continuations (and that is the correct way to go).

Actually, here's everything we need to know exactly what request is completed:
https://github.com/microsoft/FASTER/blob/bd5ad88df06e881559f56fd2b3c085e12f3229bf/cs/src/core/Allocator/AllocatorBase.cs#L1449

Could we refactor some code,  so async continuations get properly wired and only complete the corresponding read request, rather than competing for shared state?

ThiagoT1:
> Important: 
> 1) Sync methods behavior and API are unchanged
> 2) The only flow control changed is ReadAsync. Other async methods are unchanged.
> 

Properly Wire ReadAsync Continuations.

Every commit tries to stage a step towards  the PR objective.

The main point of concern is the RelaxedCPR treatment on cs/src/core/Index/FASTER/FASTER.cs

Fixes #246


ThiagoT1:
Last commit done in attempt to meet FASTER concurrency model requirements.

All tests passing, including comprehensive ones on dependent libs, like FasterDictionary:

image

image

cs/src/core/Index/FASTER/FASTERThread.cs
```diff
@@ -465,66 +440,40 @@ internal async ValueTask<(Status, Output)> CompleteIOPendingReadRequestsAsync(lo
...
 
-                OperationStatus internalStatus;
-                // Issue the continue command
-                if (pendingContext.type == OperationType.READ)
-                {
-                    internalStatus = InternalContinuePendingRead(opCtx, request, ref pendingContext, currentCtx);
-                }
-                else
-                {
-                    internalStatus = InternalContinuePendingRMW(opCtx, request, ref pendingContext, currentCtx); ;
```
badrishc: We need to support pending RMW operations. Why is this deleted? RMW operations may either "retry" or go "pending".

cs/src/core/Allocator/AsyncIOContext.cs
```diff
@@ -10,8 +11,15 @@ namespace FASTER.core
     /// <summary>
     /// Async IO context for PMM
     /// </summary>
-    public unsafe struct AsyncIOContext<Key, Value>
+    public class AsyncIOContext<Key, Value> : FasterAsyncOperation<int>, IDisposable
```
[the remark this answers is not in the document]
ThiagoT1: You're right.
We could revert it to struct and inside it create a new field to store the ```FasterAsyncOperation<int>```

badrishc:
I am worried about the depth of changes in this PR as they break subtle invariants in the system. We will step back and (1) fix the multi-async bug; (2) make sure your use case gets high performance. I think these goals can be accomplished without modifying the mono-threaded session access invariant.

cs/src/core/Index/FASTER/FASTERThread.cs
ThiagoT1: This code is only hit be readyAsync execution path.
Neither RWM or RMWAsync hit this, and all RMW (and other write operations) are unchanged.

badrishc:
Also, FASTER is a latch-free library. We try to not take locks anywhere, except in one case during checkpointing, when forcing a session to move from one version to another. This is achieved by ensuring that only the mono-threaded session accesses/updates any state.

Thus, we need the async continuations to simply be added to the async response queue (which we plan to replace with a latch-free implementation in future), from which the mono-threaded session will remove elements.

ThiagoT1:
> 
> 
> This is achieved by ensuring that only the mono-threaded session accesses/updates any state.

if this is about     FASTER.cs:389 ```lock (clientSession.ctx)```, it was added since Native32.Read callbacks do not get executed in a mono-threaded way. And the continuation in the PR is following the same execution path.
Thus, such code will also be executed in a multi threaded way, hence the lock, that by the way also happens inside ConcurrentQuere, albeit by way of ```Interlocked``` methods and spinners.


> 
> Thus, we need the async continuations to simply be added to the async response queue (which already uses a latch-free ConcurrentQueue implementation), from which the mono-threaded session will remove elements.

This would have ate least these consequences:

1) The AsyncQueue only releases a single awaiter to go, every time an enqueue happens, so
2) Even if this awaiter completed someone elses continuation and stored it somewhere, the actual awaiter would still be hung on DequeueAsync (maybe forever)
3) To counter that, we would need a Timeout (lots of allocation, bcz its a task array with two tasks in it, at least)
4) other option would be to Re-Enqueue the continuation on AsyncQueue. Then we could have endless racing between several awaiters, plus a new Task and StateMachine every loop.

In the end, we would need to use other type of queue, that would signal (by a task) every awaiter, so they can all just discard a continuation thats not theirs and re await for the next completion, until theirs come around, and they can return.

That would produce a task for every completion multiplied by maybe a new state machine for every awaiter loop interation.

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -319,6 +319,104 @@ internal Status ContextRead(ref Key key, ref Input input, ref Output output, Con
...
+                    clientSession.ctx.pendingReads.Add(pendingContext.id, pendingContext);
+
+                    token.ThrowIfCancellationRequested();
+
+                    if (@this.epoch.ThisInstanceProtected())
+                        throw new NotSupportedException("Async operations not supported over protected epoch");
+
+                    await diskRequest.ValueTaskOfT;
+
+                    lock (clientSession.ctx)
```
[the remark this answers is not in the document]
ThiagoT1: > 
> 
> What if the user does RMWAsync followed by ReadAsync with waitForCommit true. We would require the older RMW to complete as well, as part of this wait for commit.

That should be ensured by ```ClientSession.WaitForCommitAsync``` (that wen unchanged), no?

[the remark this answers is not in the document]
badrishc: If user does this sequence, the expectation is that all operations until that point are committed:

s.RMWAsync();
await s.ReadAsync(waitForCommit: true);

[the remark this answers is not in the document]
ThiagoT1: > 
> 
> We do not want to take the overhead of locking the client session. This is the reason FASTER is mono-threaded. All work for a session is done on a single thread, and this PR breaks that contract.

I see. 

However, currently there's other places were there's shared memory, which current synchronization techniques are equivalent to locks, even with the possibility of a context switch. Those are ConcurrentDictionaries and  ConcurrentQueues.

Would that be better if we switched the ```lock``` to ```Interlocked``` calls and a Spinner?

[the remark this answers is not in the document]
ThiagoT1: > 
> 
> If user does this sequence, the expectation is that all operations until that point are committed:
> 
> s.RMWAsync();
> await s.ReadAsync(waitForCommit: true);

This is happening.
RMWAsync will put an entry inside the pending requests collection, that in turn will be monitored by the ```CompletePendingAsync``` call inside ```WaitForCommitAsync```

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -319,6 +319,108 @@ internal Status ContextRead(ref Key key, ref Input input, ref Output output, Con
...
+
+            var nextSerialNum = clientSession.ctx.serialNum + 1;
+
+            OperationStatus internalStatus;
+
+            if (clientSession.SupportAsync) clientSession.UnsafeResumeThread();
+            try
+            {
+                do
+                {
```
badrishc: This loop is not identical in behavior to the original code. For example, I notice that InternalRefresh would not be called in this tight loop, resulting in never getting out of CPR. The debug asserts for version correctness are also skipped. We need to ensure that all code paths are functionally identical to the older case.

Why not use an async equivalent of HandleOperationStatus? This keeps the main FasterKV.ReadAsync method small so it can be inlined, leaving the complexity of uncommon cases to a sub-method.

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -319,6 +319,108 @@ internal Status ContextRead(ref Key key, ref Input input, ref Output output, Con
...
+                    diskRequest = await diskRequest.asyncOperation.ValueTaskOfT;
+
+                    lock (clientSession.ctx)
+                    {
+
+                        if (clientSession.SupportAsync) clientSession.UnsafeResumeThread();
+                        try
+                        {
+                            #region RelaxedCPR
+                            if (!@this.RelaxedCPR)
```
badrishc: For async code ONLY: we can assume and assert that RelaxedCPR is true. This should help simplify the code (including getting git of release of latch in InternalCompleteIOPendingReadRequestAsync).

ThiagoT1: Code replaced with a ```Debug.Assert(@this.RelaxedCPR);```

ThiagoT1:
As per last discussions, a new struct is wrapping the compute completion logic, so the user can trigger it.
The completion lock is gone and the user must call ```ReadAsyncResult.CompleteRead``` in a single threaded manner.

ThiagoT1:
BTW,

Thanks you Mr. Badrish for the patience and the high level discussions about this issue.

cs/src/core/Allocator/AsyncIOContext.cs
```diff
@@ -52,6 +54,12 @@ public unsafe struct AsyncIOContext<Key, Value>
         /// </summary>
         public AsyncQueue<AsyncIOContext<Key, Value>> callbackQueue;
 
+
+        /// <summary>
+        /// Async Operation ValueTask backer
+        /// </summary>
+        public FasterAsyncOperation<AsyncIOContext<Key, Value>> asyncOperation;
```
badrishc: Is there a good reason to use all the complicated logic in FasterAsyncOperation vs a TaskCompletionSource that we set completed when IO is done?

badrishc: We should avoid introducing new code if we can avoid it.

cs/src/core/Index/FASTER/FASTERImpl.cs
```diff
@@ -1842,6 +1836,40 @@ internal Status InternalContainsKeyInMemory(ref Key key, FasterExecutionContext
...
+            Debug.Assert(currentCtx.version == version + 1);
+            Debug.Assert(currentCtx.phase == Phase.IN_PROGRESS);
+
+            pendingContext.version = currentCtx.version;
+        }
+
+        internal AsyncIOContext<Key, Value> ScheduleGetFromDisk(FasterExecutionContext opCtx,
+                    ref PendingContext pendingContext)
+        {
+            pendingContext.id = opCtx.serialNum;
```
badrishc: This results in a id of -1 in the beginning. Just fixing it to serialNum+1 is also not desirable because in future we will allow multiple ReadAsync operations to have the same serialNum. Therefore we have to use the `totalPending` field we had earlier to ensure ids are unique: `pendingContext.id = opCtx.totalPending++;`

Related, can you reconsider adding back the use of `HandleOperationStatus`? That will allow code to not diverge between the sync and async versions? If not, that is fine and we will take care of it after this PR. Right now, there is duplication of logic in two places which makes maintenance a nightmare.

Every line and variable in the original code is there for a reason, so please go through to see if any other issues like the above may have cropped up.

cs/src/core/Index/FASTER/FASTERThread.cs
```diff
@@ -105,6 +105,7 @@ internal void InitContext(FasterExecutionContext ctx, string token, long lsn = -
                     ctx.retryRequests = new Queue<PendingContext>();
                     ctx.readyResponses = new AsyncQueue<AsyncIOContext<Key, Value>>();
                     ctx.ioPendingRequests = new Dictionary<long, PendingContext>();
+                    ctx.pendingReads = new AsyncCountDown<long, PendingContext>();
```
badrishc: Do we still need this? Why not just use ioPendingRequests as before?

badrishc: All we need now is a counter, no need for a dictionary, correct?

cs/src/core/Allocator/AsyncIOContext.cs
ThiagoT1: Yes.
TCS will yield a task (class) while our FAO will yield and back a ValueTask (struct).
Its one allocation less.

Lots of code inside .net core itself are using IValueTaskSource as is FasterAsyncOperation, as to avoid allocating a Task when possible.

cs/src/core/Index/FASTER/FASTERThread.cs
ThiagoT1: Former logic would have ```CompleteIOPendingRequestsAsync``` loop, await and carry the completion of every pending request. Until all pending things are done with.

New logic (thankfully) does not hijack control to do other continuation's business. So, a way to await for the pending count to reach zero is needed.

Also, the counter is a dictionary bcz there's other paths where every element of the pending collection will be traversed:

```csharp
foreach (var _ctx in ctx.pendingReads)
{
    AcquireSharedLatch(_ctx.Value.key.Get());
}
```

cs/src/core/Index/FASTER/FASTERImpl.cs
ThiagoT1: > This results in a id of -1 in the beginning. Just fixing it to serialNum+1 is also not desirable because in future we will allow multiple ReadAsync operations to have the same serialNum. Therefore we have to use the totalPending field we had earlier to ensure ids are unique: pendingContext.id = opCtx.totalPending++;

Fixed by last commit.

cs/src/core/Index/FASTER/FASTERThread.cs
ThiagoT1: Also, I foresaw the lock inside ```AsyncCountDown``` would raise concern, so I made it lock free.

Issue #244, opened 2020-02-14
[C#] FasterLogScanIterator.GetNext should return address of next entry


mito-csod:
With the update in 2020.2.1.1, the out long address argument got changed to contain the current address--this renders FasterLogScanIterator.CompleteUntil() unusable, since it expects the "next" address as its argument, but there is no thread-safe way to get this value outside of the previous implementation of out long address. Please consider either reverting the behavior, or adding another out argument for the next address.

Thanks.

marius-klimantavicius:
Checkout `FasterLogScanIterator.NextAddress` field.

mito-csod:
I think that falls in the "not thread safe" category.

Imagine if you have multiple threads calling into GetNext() of a single iterator so that they can do CPU-heavy processing on the data in parallel. Unless you use some synchronization (like a mutex), you have no guarantee that another thread hasn't already called GetNext() in between the time you call your GetNext() and the time you fetch NextAddress. While we certainly *can* do that, I think the most elegant solution would be to return the "would-be-next address" at the call of GetNext() so that we avoid that overhead.

badrishc:
Agreed, this was discussed on Gitter earlier. We need it to be thread-safe, and the call stack is the best way to ensure this.

badrishc:
This requires a bit more clarification. Assume we have two threads calling GetNext simultaneously. A thread calls GetNext, does some processing on the returned item, then calls GetNext again in a loop. Now consider from the perspective of one of these threads.

When it calls GetNext and gets the "next" address (via the proposed fix), the next address represents the address of the entry immediately after the currently returned entry. It does NOT imply that the other concurrent thread has completed processing any previous entry. In other words, FASTER does not know when an iterator thread has finished processing its retrieved entry.

That said, the above proposed extension to get the immediate next address might still be more useful than calling iter.NextAddress - because the former would tell the caller the exact address of the item immediately after the one currently retrieved, whereas the latter might return an address further into the future.

PR #251, opened 2020-03-02
Added overload of GetNext() that also returns next address

mito-csod:
Here is a proposed overload of FasterLogIterator.GetNext which also has an _out nextAddress_ with a simple test case. GetNextInternal was also updated for it.

Sending for your consideration. I also am not a regular user of GitHub, so apologies if this comes out wrong.

Thanks!

Resolves microsoft/FASTER#244

### DIFF
--- a/cs/src/core/Index/FasterLog/FasterLogIterator.cs
+++ b/cs/src/core/Index/FasterLog/FasterLogIterator.cs
@@ -512,8 +512,8 @@ namespace FASTER.core
 
                 if (Utility.MonotonicUpdate(ref NextAddress, currentAddress, out long oldCurrentAddress))
                 {
+                    nextAddress = currentAddress;
                     currentAddress = oldCurrentAddress;
-                    nextAddress = NextAddress;
                     return true;
                 }
             }
--- a/cs/src/core/Index/FasterLog/FasterLogIterator.cs
+++ b/cs/src/core/Index/FasterLog/FasterLogIterator.cs
@@ -185,8 +185,21 @@ namespace FASTER.core
         /// <returns></returns>
         public unsafe bool GetNext(out byte[] entry, out int entryLength, out long currentAddress)
         {
+            return GetNext(out entry, out entryLength, out currentAddress, out _);
+        }
+
+        /// <summary>
+        /// Get next record in iterator
+        /// </summary>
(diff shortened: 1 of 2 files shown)


badrishc:
Thanks for the contribution. Your changes look great. Once you sign the CLA (see above), we should be able to accept it. Thanks!

cs/src/core/Index/FasterLog/FasterLogIterator.cs
```diff
@@ -484,11 +513,10 @@ private unsafe bool GetNextInternal(out long physicalAddress, out int entryLengt
                 if (Utility.MonotonicUpdate(ref NextAddress, currentAddress, out long oldCurrentAddress))
                 {
                     currentAddress = oldCurrentAddress;
+                    nextAddress = NextAddress;
```
marius-klimantavicius: Not sure if this code can be run by multiple threads simultaneously, but if it can then maybe: 
`nextAddress = currentAddress;`

badrishc: Ah yes, you are right.

Otherwise it defeats the purpose of adding this extra parameter, since one could just check NextAddress as you mentioned in your message earlier.

badrishc: Just so everyone gets the context, the situation this catches is that there are two GetNext calls in parallel. The caller can safely finalize only until the immediate next address _for their own GetNext_, because from their perspective they have completed the processing only of the row they just retrieved.

badrishc: This above argument does assume that when a caller finishes processing a row, they somehow know that all previous rows have also completed processing, so it's safe to finalize until the end of that address.

badrishc: See here for a discussion: https://github.com/microsoft/FASTER/issues/244#issuecomment-593519144

mito-csod: Good catch. I will update the PR as suggested.
Re: the "completed up to" situation, the reading entity probably needs to hold on to both the current and the next addresses in some synchronized collection(s), and only issue a CompleteUntil() up to the point all known nexts are accounted for by the known currents.

PR #269, opened 2020-05-22
Add ReadyToCompletePendingAsync for client to check if there is somet…

badrishc:
…hing to complete-pending on.

### DIFF
--- a/cs/src/core/ClientSession/FASTERAsync.cs
+++ b/cs/src/core/ClientSession/FASTERAsync.cs
@@ -28,7 +28,7 @@ namespace FASTER.core
     {
 
         /// <summary>
-        /// Check if at least one request is ready for CompletePending to operate on
+        /// Check if at least one (sync) request is ready for CompletePending to operate on
         /// </summary>
         /// <param name="clientSession"></param>
         /// <param name="token"></param>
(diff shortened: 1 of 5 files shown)


cs/src/core/ClientSession/FASTERAsync.cs
```diff
@@ -26,6 +26,30 @@ public partial class FasterKV<Key, Value, Input, Output, Context, Functions> : F
...
+        /// <returns></returns>
+        internal async ValueTask ReadyToCompletePendingAsync(ClientSession<Key, Value, Input, Output, Context, Functions> clientSession, CancellationToken token = default)
+        {
+            #region Previous pending requests
+            if (!RelaxedCPR)
+            {
+                if (clientSession.ctx.phase == Phase.IN_PROGRESS || clientSession.ctx.phase == Phase.WAIT_PENDING)
+                {
+                    if (clientSession.ctx.prevCtx.ioPendingRequests.Count != 0)
+                        await clientSession.ctx.prevCtx.readyResponses.WaitForEntryAsync();
```
sebastianburckhardt, line 44: probably want to pass the cancellation token as an argument here?

cs/src/core/ClientSession/FASTERAsync.cs
```diff
@@ -26,6 +26,30 @@ public partial class FasterKV<Key, Value, Input, Output, Context, Functions> : F
...
+                }
+            }
+            #endregion
+
+            if (clientSession.ctx.ioPendingRequests.Count != 0)
+                await clientSession.ctx.readyResponses.WaitForEntryAsync();
```
sebastianburckhardt, line 50: same here

sebastianburckhardt:
I have tested this feature in our environment, it seems to work as intended.

badrishc:
Fixed based on comments.

Issue #230, opened 2020-01-10
[C#] Proposal: Let sessions provide implementation of IFunctions (callbacks)


badrishc:
Right now, the FASTER constructor takes `IFunctions` as input. These function callbacks identify how to perform operations such as reading/writing, RMW in-place and copy-updates, etc.

An interesting alternative would be to provide them later, during session creation, and associate the functions with sessions instead of the main FASTER instance. This would let different sessions use different functions. E.g., one session could perform RMW for incrementing field1 while another could perform a different function.

It would make it easier to use the same FASTER instance to store different forms of data (one can do it even today, via functions that switch between the different operations based on input, but maybe a bit more cumbersome). Note that all the sessions will still see the same data types (TKey, TValue, TInput, etc.) but will be able to operate on the data in a customized way.

Further, the late binding of functions means that users can change the function being performed during runtime, by simply deleting the older session and creating a new session with the new functions.

There might be a performance implication due to the late binding - this is as yet unknown.

dlandi:
I like it.

ThiagoT1:
I'm doing exactly this right know.

When effectivelly trying to save different data types on the same KV, the serializers get funny.

One way to work around this would be to provide serializers to a KVSession as well.

This would imply that TKey and TValue would move to KVSession, and leave FasterKV, right?

badrishc:
The proposal was to leave the TKey and TValue types in FasterKV. The client sessions only get to define a different set of functions on the same types. You would have to typecast/use inheritance if you wanted to use different types.

ThiagoT1:
That would likely make this Functions reference become an interface:

https://github.com/microsoft/FASTER/blob/bd5ad88df06e881559f56fd2b3c085e12f3229bf/cs/src/core/Index/FASTER/FASTER.cs#L20

Is that OK?

On that note, for how long must ```FasterKV``` thread affinity be supported? 
If we ditch it right away, in favor of ```ClientSession```, the code would get a log simpler.

marius-klimantavicius:
I have a PoC of this in https://github.com/marius-klimantavicius/mister/tree/functions_in_session (inside `FASTER.core` folder). Haven't had time (will not have for the near future) to run any tests though.
 
[...]

PR #270, opened 2020-05-27
Functions in session

marius-klimantavicius:
For https://github.com/microsoft/FASTER/issues/230.
Not sure about naming (`ISynchronizationListener`).
There is at least one breaking change (`ReadAsyncResult` is in different class [different number of type arguments]).
For now I duplicated state machine code for sync and async versions, if this is not ideal I can unify into one method again.
Needed to introduce `IClientSession`, but this seems to be one-time call so no harm there.


marius-klimantavicius:
Fixed naming (at least to my liking).
Still missing tests and example.

badrishc:
Thanks for the contribution. As this is a breaking change, we need to better understand what benefits it brings users, in order to justify the change.

@tli2 what do you think of this: "For now I duplicated state machine code for sync and async versions, if this is not ideal I can unify into one method again."

tli2:
@badrishc 

I only took a cursory look. In principle I think that's the right thing to do. For this particular case there also seems to be not too much actual code duplication.

If I am to be nit-picky I'd say we need a better model for when it is appropriate to use which, and reflect that in the API naming and documentation. We might also consider enforcing that separation in the code (e.g. if we decide that `OnThreadAsync` can only be called from sessions that support async, we can check that in FasterStateMachine.cs)

marius-klimantavicius:
Should I experiment with moving `Input`, `Output` and `Context` out of `FasterKV` (to be part of method signature not of class itself, part of `IFunctions` and `ClientSession`)? Rationale - different callbacks might work with different `Intput`, `Output`, `Context`.

marius-klimantavicius:
One (relatively annoying) issue with moving `Input`, `Output` and `Context` is that type inference for `NewSession` and `ResumeSession` do not work for primary `FasterKV<Key, Value>` (there is also legacy `FasterKV` for backwards compat).

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -241,27 +241,28 @@ public ValueTask<long> EnqueueAsync(byte[] entry, CancellationToken token = defa
                 return new ValueTask<long>(logicalAddress);
 
             return SlowEnqueueAsync(this, entry, token);
-            static async ValueTask<long> SlowEnqueueAsync(FasterLog @this, byte[] entry, CancellationToken token)
+        }
+
+        private static async ValueTask<long> SlowEnqueueAsync(FasterLog @this, byte[] entry, CancellationToken token)
```
badrishc, line 246: Any reason for this code to change?

cs/src/core/Index/Common/AddressInfo.cs
```diff
@@ -74,7 +74,7 @@ public long Size
 
         public long Address
         {
-            readonly get
+            get
```
badrishc, line 77: why is readonly marker removed?

cs/src/core/Index/FasterLog/FasterLog.cs
marius-klimantavicius: I am embedding FASTER as source code into net462 project, so this is one of the places where I have to manually "fix" to be able to compile and there is no difference from performance standpoint (one can also say - personal preference).

cs/src/core/Index/Common/AddressInfo.cs
marius-klimantavicius: Remnant from embedding source code to net462 - will re-add this back.

badrishc:
I think it would be great for us to support net 462 as well, so those changes are welcome, I would think, as long as perf is unaffected.

marius-klimantavicius:
Latest commit is for #272 - Added custom compaction functions. Note that in addition to having custom `IsDeleted` I added `Copy` and `CopyInPlace` to be able to deal with concurrency (now that I am writing this - do we really need these? Is it possible to compact *inside* mutable region? Probably yes as `untilAddress` can be tail).

Net462 is supported when using FASTER as binary dependency (e.g. via nuget) however I am embedding sources directly (in addition I sometimes do have more changes) thus locally I've removed `readonly` members, nullables, local static functions, etc. For PRs back to FASTER I am trying to keep it close to original but still some changes might come through (like extracting static local functions to full functions [this is more of a personal preference], I can move them back).

marius-klimantavicius:
I think I will do one more change - I want to add an overload to `IVariableLengthStruct.GetLength` to take in `TInput`. It is possible that due to RMW I might want to change the size of the record based on the incoming `input`.

badrishc:
> For now I duplicated state machine code for sync and async versions, if this is not ideal I can unify into one method again.

I think this PR should not duplicate the async and sync versions, and instead should keep them unified as before. It seems orthogonal to the intent of this PR, correct me if that's not the case. We can consider the issue of better handling the async vs sync state machine in a separate PR.

badrishc:
Hi @marius-klimantavicius, this is a really nicely done PR, thanks for your contribution. I am done with the review and have no other comments than this one. Let me know what you think.

> For now I duplicated state machine code for sync and async versions, if this is not ideal I can unify into one method again.

I think this PR should not duplicate the async and sync versions, and instead should keep them unified as before. It seems orthogonal to the intent of this PR, correct me if that's not the case. We can consider the issue of better handling the async vs sync state machine in a separate PR.

badrishc:
One thing left is to verify that there is no performance regression - have you done some testing in this regard?

@TedHartMS - can we use your PerfTest benchmark PR to see the impact of this PR?

marius-klimantavicius:
Removed sync methods from state machine (it seems that during my experiments I stopped using sync version after all).

As for performance regression - I haven't performed any testing regarding this. Later will try to see if I can use PerfTest for that.

marius-klimantavicius:
I tried running PerfTest a few times (I can't find YCSB files so I am using synthetic test only) however I am unable to get stable results on my machine (even with fixed seed for random number generator). I've tried running only a few times (~10-15) as each run (Release build, Ctrl-F5) takes at least 5 minutes.

TedHartMS:
What's your machine configuration? The Benchmark test's Release build has a working set around 24GB. The PerfTest app @badrishc referred to is a separate branch and will be available soon.

marius-klimantavicius:
I see, I have only 16GB of RAM on this machine maybe it was due swapping (I might try the PerfTest branch a bit later).

badrishc:
We are noticing a 10% drop in performance on a single thread.

## Before

FASTER.benchmark.exe -b 0 -t 1 -n 1 -r 50 -d zipf

loading keys from D:\ycsb_files\load_zipf_250M_raw.dat into memory...
loaded 250000000 keys.
loading txns from D:\ycsb_files\run_zipf_250M_1000M_raw.dat into memory...
loaded 1000000000 txns.
Executing setup.
**Loading time: 79089ms**
Executing experiment.
Thread 0 done; 50645771 reads, 50645996 writes, in 30105 ms.
End tail address = 6000001496
Total 101291767 ops done  in 30.107 secs.
##, zipf, 1, 50, 1, **3364392.5665127714**, 0


## After

FASTER.benchmark.exe -b 0 -t 1 -n 1 -r 50 -d zipf
loading keys from D:\ycsb_files\load_zipf_250M_raw.dat into memory...
loaded 250000000 keys.
loading txns from D:\ycsb_files\run_zipf_250M_1000M_raw.dat into memory...
loaded 1000000000 txns.
Executing setup.
**Loading time: 88491ms**
Executing experiment.
Thread 0 done; 45542661 reads, 45542155 writes, in 30139 ms.
End tail address = 6000001496
Total 91084816 ops done  in 30.141 secs.
##, zipf, 1, 50, 1, **3021957.333864172**, 0

badrishc:
With a minor update to ClientSession.cs, I was able to restore performance to the same level as before. Please review if the change is ok with you.

marius-klimantavicius:
Yep these are absolutely OK (I was hoping `new AsyncFasterSession(this)` to be optimized away but I was wrong). Having `readonly functions` is the same as before this PR (though once I figure out how to properly benchmark [provided my pc is powerful enough] I will check if perf is not affected when `functions` is `struct` [due to defensive copy though only matter for non-empty structs]).

badrishc:
Yeah, 'new AsyncFasterSession(this)' is the main change that got back the 10%. The other change was just a move to be in the safe side. Thanks again!

badrishc:
I tried removing and re-adding the `readonly` marker for functions, but the results are inconclusive. I'm seeing +/- 10% across runs. It does not block this PR from getting merged, but it would be nice to get to the bottom of whats going on in the critical path, and why certain benchmark runs are regularly (but not always) 10% slower.

badrishc:
> One (relatively annoying) issue with moving `Input`, `Output` and `Context` is that type inference for `NewSession` and `ResumeSession` do not work for primary `FasterKV<Key, Value>` (there is also legacy `FasterKV` for backwards compat).
> 
> Can slightly reduce the noise by adding helper "builder" (still need to specify `Input`, `Output` and `Context` but functions can be inferred):
> 
> ```
> Builder<Input, Output, Context> For<Input, Output, Context>() => new Builder<...>(this);
> ```
> 
> ```
> using (var session = fasterKV.For<MyInput, MyOutput, MyContext>().NewSession(new MyFunctions()...)
> ```

@marius-klimantavicius: In PR https://github.com/microsoft/FASTER/pull/310, I have extended the builder slightly, to support writing of slightly cleaner code that produces a proper client session with the right generic params (i.e., specific `Function` instead of `IFunctions<>`):

```cs
store.For(new MyFunctions()).NewSession<MyFunctions>(...);
```

It is still ugly (`MyFunctions` has to be uttered twice), but at least user does not need to utter the types of `Input`, `Output`, `Context`.

PR #273, opened 2020-06-03
Fast-forward client session trivially to the current system state, if…

badrishc:
… the current system state were not associated with any state machine (null currentSyncStateMachine implies a trivial state machine). This can happen after recovery.


badrishc:
Pipeline build break will be fixed after this PR rolls out: [microsoft/azure-pipelines-tasks#13058](https://github.com/microsoft/azure-pipelines-tasks/pull/13058)

Issue #253, opened 2020-03-07
[C#]Enumerable FasterKV


qinxg:
Hello
I want to Enumerable FasterKV key and value   .  
Like Dictionary Foreach 
How to achieve?

Need Exclude historical data

`var iter = db._db.Log.Scan(db._db.Log.BeginAddress, db._db.Log.TailAddress, ScanBufferingMode.SinglePageBuffering);
            while (iter.GetNext(out RecordInfo recordInfo, out key, out value))
            {
            }`

ThiagoT1:
Currently, you have to use the log scan logic.

Sample here:
```csharp
_iterationState = new IterationContextState()
{
    Functions = new Functions(_options.Logger),
    UntilAddress = KV.Log.TailAddress
};

var variableLengthStructSettings = new VariableLengthStructSettings<VariableEnvelope, VariableEnvelope>()
{
    keyLength = VariableEnvelope.Settings,
    valueLength = VariableEnvelope.Settings
};

_iterationState.TempKV = new FasterKV<VariableEnvelope, VariableEnvelope, byte[], byte[], Context, Functions>
    (KV.IndexSize, _iterationState.Functions, new LogSettings(), comparer: KV.Comparer, variableLengthStructSettings: variableLengthStructSettings);

_iterationState.KVSession = _iterationState.TempKV.NewSession();

using (var iter1 = KV.Log.Scan(KV.Log.BeginAddress, KV.Log.TailAddress))
{

    while (iter1.GetNext(out RecordInfo recordInfo))
    {
        ref var key = ref iter1.GetKey();
        ref var value = ref iter1.GetValue();

        _iterationState.KVSession.Delete(ref key, default, 0);

        if (!recordInfo.Tombstone)
            _iterationState.KVSession.Upsert(ref key, ref value, default, 0);
    }
}

// TODO: Scan until SafeReadOnlyAddress
long scanUntil = _iterationState.UntilAddress;
LogScanForValidity(ref _iterationState.UntilAddress, ref scanUntil, ref _iterationState.KVSession);

// Make sure key wasn't inserted between SafeReadOnlyAddress and TailAddress

_iterationState.KeyIteration = _iterationState.TempKV.Log.Scan
(
    _iterationState.TempKV.Log.BeginAddress,
    _iterationState.TempKV.Log.TailAddress
);
```

[...]

Issue #181, opened 2019-10-01
Enumerate Keys


ThiagoT1:
First of all, thanks a lot for the amazing work done here.

Is there any recommended strategy on how to iterate over all the keys from a given collection?

In absence of that, I'm guessing some Segmented Index must be maintained by the user? 😱 

Thanks!

badrishc:
You can iterate over the log using our log scan interface (https://github.com/microsoft/FASTER/pull/90). This would iterate through all entries in the log. You might see a key more than once if it was subsequently updated on the log (i.e., when you updated the key, it was no longer in the mutable region and therefore had to be copied to tail).

To get exact iterator semantics (read only current value of each key), you need to scan the log from head to tail and build an in-memory hash table (using FASTER) of the latest values of keys. Then you can simply iterate through this in-memory hash table's entries. We do something very similar in our log compaction code here: https://github.com/microsoft/FASTER/blob/f3c3eb000c2a8f43fa0d8a3aee315cc47087550e/cs/src/core/Index/FASTER/LogAccessor.cs#L183

It would be interesting and useful to expose this as a general iterator functionality on FASTER.

PR #287, opened 2020-07-14
[C#] [WIP] Iterator for all distinct keys in FASTER KV

badrishc:
Proper solution for https://github.com/microsoft/FASTER/issues/253
Low memory footprint in common case of rare hash collisions


badrishc:
Another related issue: https://github.com/microsoft/FASTER/issues/181

Issue #289, opened 2020-07-23
[C#] Input-specific functions in IVariableLengthStruct


badrishc:
In https://github.com/microsoft/FASTER/pull/270, we added an additional method to `IVariableLengthStruct`:

`int GetLength<Input>(ref T t, ref Input input)`

In addition, we already had this earlier:

`int GetInitialLength<Input>(ref Input input)`

With the moving of `Input` to client sessions (which is a huge improvement, btw), it is causing complications when users have sessions with different input types on the same FASTER store, as the above two functions cannot be typed specifically to a session.

The proposal would be to instead provide these two methods via `ClientSession` (internally, `FasterSession`) so that they may be customized per Input type.

CC @marius-klimantavicius as you authored the above-referenced PR and are in the best position to comment/contribute on this.

badrishc:
Also cc @mito-csod

marius-klimantavicius:
We could move out methods from `IVariableLengthStruct` to `IFasterSession<...>` what I am thinking about is how to handle `Generic` and `Blittable` cases (relatively easy to do) and how to allow the user to specify the required methods (without affecting fixed size cases). One idea is to have `IVariableLengthStruct` as is (or with `<Input>` methods removed) and introduce additional `ISessionVariableLengthStruct` (or similar, naming is difficult) which could be optionally provided when creating client session (for fixed sizes it would be either exception or not used).

badrishc:
Mates sense. We could introduce `IVariableLengthStruct<T, TInput>` that could be provided per session. If not provided, we would default to fall through to existing `IVariableLengthStruct`.

badrishc:
PR is out, will merge soon based on feedback if any.

PR #290, opened 2020-07-28
Separate out Input-dependent functions by session, for variable length structs

badrishc:
* Allow users to specify var-len struct settings that are Input-dependent, on a per-session basis, using a new IVariableLengthStruct<T, Input>.
* Remove Input-based methods from IVariableLengthStruct<T> interface.
* Provide default implementation for the Input-dependent functions
GetLenth(Input) -> GetAverageLength()
GetLength(Value, Input) -> GetLength(Value)

Fix https://github.com/microsoft/FASTER/issues/289

### DIFF
--- a/cs/src/core/ClientSession/ClientSession.cs
+++ b/cs/src/core/ClientSession/ClientSession.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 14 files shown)


cs/src/core/ClientSession/ClientSession.cs
```diff
@@ -32,22 +33,39 @@ public sealed class ClientSession<Key, Value, Input, Output, Context, Functions>
...
+            {
+                if (fht.hlog is VariableLengthBlittableAllocator<Key, Value> allocator)
+                {
+                    Debug.WriteLine("Warning: Session did not specify Input-specific functions for variable-length values via IVariableLengthStruct<Value, Input>");
+                    this.variableLengthStruct = new DefaultVariableLengthStruct<Value, Input>(allocator.ValueLength);
+                }
+            }
+            else
+            {
+                if (fht.hlog is VariableLengthBlittableAllocator<Key, Value>)
```
marius-klimantavicius, line 65: Probably `!(fht.hlog is ...)`

badrishc: Fixed, thx.

PR #293, opened 2020-08-06
Yield thread when spinning on Intermediate state

tli2:
In some cases when there are more threads than physical cores, the thread responsible for transition can be scheduled out while others busy-spin on the state, leading to a significant drop in performance. Calling `Thread.Yield()` when spinning reduces that overhead.


tli2:
Investigating if a semaphore is better than spinning

PR #297, opened 2020-08-11
[C#] IDevice v2

badrishc:
* In service of new scenarios involving extremely high performance devices as well as our tiered and sharded meta-devices, we are moving to a cleaner `IDevice` abstraction that no longer involves `NativeOverlapped*`. The latter concept is very particular to Windows local disk drives, and creates unnecessary overhead when used with other devices.

* We also found that `Overlapped` creation in C# does not scale beyond 4 or 8 cores for NVMe drives. This PR fixes the problem by pooling `Overlapped` objects and minimizing their creation overhead. For a 100% random disk read workload, we now get scalability within 20% of the NVMe disk IOPS (up to 600K IOPS for our device) using 8 cores and `LocalStorageDevice` in Windows.

* We have also added a new high-performance `LocalMemoryDevice` that stores the data in memory, for easier testing of high performance devices. With this device, for a uniform random workload with all operations going to the local memory device, we get around 2.6 million ops/sec. We use a sector size of 64 bytes for main memory, as that corresponds to cache size. This is using 4 FASTER threads and 1 async I/O thread.

* The notion of _device throtting_ during reads (max `numPendingReads` of 120) has now been made a capability of individual devices via a `Throttle` API, instead of hard-coding it at the FASTER level.

* Some improvements to `ManagedLocalStorageDevice` are incorporated as well, to avoid memory copies in netstandard21.

* Use `HandleCapacity` in `StorageDeviceBase` to enable individual devices respect max capacity.

Thanks  @t-qizzh for collaboration on this PR.


cs/src/core/Device/LocalMemoryDevice.cs
```diff
@@ -0,0 +1,182 @@
...
+                var x = i;
+                ioQueue[x] = new ConcurrentQueue<IORequestLocalMemory>();
+                ioProcessors[i] = new Thread(() => this.ProcessIOQueue(ioQueue[x]));
+                ioProcessors[i].Start();
+            }
+
+            Console.WriteLine("LocalMemoryDevice: " + ram_segments.Length + " pinned in-memory segments created, each with " + sz_segment + " bytes");
+        }
+
+        private void ProcessIOQueue(ConcurrentQueue<IORequestLocalMemory> q)
```
hiteshmadan, line 72: Superlike the idea of a fully supported memory device!

The tradeoff needed to use this device is a bit difficult to reason about: 
either add a lot of dedicated threads to minimize the queuing when reading/writing, or keep the number of dedicated IO threads low to avoid unnecessary spinning and waste of cpu%.

Is it possible to suspend the read threads until new writes show up with some basic signaling between those code paths?

Alternately, can the ConcurrentQueue be replaced with Channel<T> which has async read/write APIs? This will subject this code to the regular task scheduler, though, and cause slowdowns - maybe having a dedicated task scheduler for such hi-pri work is a good balance?

badrishc: Yes, there is a spin-wait with `Thread.Yield()` for the queue at the moment, which is sub-optimal. We could easily add a semaphore to avoid the spin, at the expense of slightly higher latency in low-load situations. We need to avoid unnecessary signaling by each enqueue, which requires some additional logic beyond what we already have in `AsyncQueue`.

Issue #255, opened 2020-03-09
[C#] UnCommit Log Scan


qinxg:
Enqueu Some Log into Fasterlog But not Commit 

Now I Scan it..   Is null...

How Can I Scan Memory Log???

![image](https://user-images.githubusercontent.com/1769735/76236591-e0269580-6267-11ea-87ed-f586ac808d6c.png)

badrishc:
The scan is by design restricted to scan only committed data because that is the normal use case for a persistent log. Users typically do not want to scan uncommitted data. What is the use case to scan uncommitted data?

As a workaround, just comment out the second part of the if condition here: https://github.com/microsoft/FASTER/blob/26569b223cdc5da62f9364c921d00a2201767bf1/cs/src/core/Index/FasterLog/FasterLogIterator.cs#L443

tpajurek-dtml:
Hello  @badrishc ,

we have also run into a use-case that requires such behavior.

We would like to use Faster Log basically as a queue between producer and consumer in the following way:

* We do not mind losing some data in case of failures (so it is good enough to commit at some intervals). 
* We expect that consumer will be faster than producer in most cases (therefore the log is usually almost empty).
* We would like to minimize writing to disk - in our case, we wouldn't need to write most of the data because they care consumed immediately after enqueuing.
* Sometimes it can happen that consumer is unable to process the data for a longer period of time and therefore the log starts to filling up. In this case, we would like to spill the data to the disk periodically.

I have tried to comment out the second part of condition as you propose but it looks like it does not work. We observe that when such code modification is done, the `GetAsyncEnumerable` returns only the first message and than blocks and does not return anymore messages if they are commited.

Is there an another way how to achieve the desired behavior?

We are using this code to test it out:

```csharp
class Program
{
    static async Task Main(string[] args)
    {
        var device = Devices.CreateLogDevice($"<path>");

        var log = new FasterLog(new FasterLogSettings { LogDevice = device });

        using var cts = new CancellationTokenSource();

        var producer = ProducerAsync(log, cts.Token);
        var commiter = CommiterAsync(log, cts.Token);
        var consumer = ConsumerAsync(log, cts.Token);

        Console.CancelKeyPress += (o, eventArgs) =>
        {
            Console.WriteLine("Cancelling program...");
            eventArgs.Cancel = true;
            cts.Cancel();
        };

        await producer;
        await consumer;
        await commiter;

        Console.WriteLine("Finished.");

    }

[...]

PR #298, opened 2020-08-17
[C#] [WIP] FasterLog support for scanning uncommitted data

badrishc:
We add the capability for scanning uncommitted data in FasterLog. You initiate a log.Scan with `scanUncommitted` set to `true` - this will allow readers to read beyond the committed address in the log. As long as readers keep up with writers, the log can stay entirely in memory and avoid being written out to disk. During transient overload, when readers are unable to keep up with writers, the logs will start getting written to disk.

Calling `Commit` will cause one I/O, as will the auto-commit at page boundaries. This is necessary to write out the current commit metadata (log begin and end address) and iterator locations, for recovery purposes. You can adjust the page size to control the frequency of auto-commit.

The tail of the log is generally not safe to consume -- multiple threads inserting into the tail at the same time means that there will be holes in the log until the threads complete their respective enqueues. To handle this, we add a method called `RefreshUncommitted` (with async variants). Similar to commit, this call exposes the tail of the log to consumers in a safe way. Since it is slightly more expensive, we do not automatically perform this operation, and instead expose to users.

You use `RefreshUncommitted` similarly to `Commit` -- either call from the enqueue thread or have a separate thread that periodically calls it.

For optimal performance, use more than one page in memory (e.g., 2 or 4 pages), and set mutable fraction to say 0.5. This will ensure that pages get auto-committed only when the in-memory log of 2 of 4 pages is 50% full. This will allow pages sufficient time for records to be consumed by readers before the commit tries to push them to disk. The example [here](https://github.com/microsoft/FASTER/tree/scan-uncommitted/cs/playground/FasterLogPubSub) show how to use this.

Fix https://github.com/microsoft/FASTER/issues/255


cs/src/core/Device/Devices.cs
```diff
@@ -24,7 +24,7 @@ public static class Devices
         /// <param name="capacity">The maximal number of bytes this storage device can accommondate, or CAPACITY_UNSPECIFIED if there is no such limit</param>
         /// <param name="recoverDevice">Whether to recover device metadata from existing files</param>
         /// <returns>Device instance</returns>
-        public static IDevice CreateLogDevice(string logPath, bool preallocateFile = true, bool deleteOnClose = false, long capacity = CAPACITY_UNSPECIFIED, bool recoverDevice = false)
+        public static IDevice CreateLogDevice(string logPath, bool preallocateFile = false, bool deleteOnClose = false, long capacity = CAPACITY_UNSPECIFIED, bool recoverDevice = false)
```
tpajurek-dtml, line 27: Breaking change.

cs/src/core/Index/FasterLog/FasterLogIterator.cs
```diff
@@ -97,43 +100,45 @@ internal unsafe FasterLogScanIterator(FasterLog fasterLog, BlittableAllocator<Em
         /// Async enumerable for iterator
         /// </summary>
         /// <returns>Entry, entry length, entry address</returns>
-        public async IAsyncEnumerable<(byte[], int, long)> GetAsyncEnumerable([EnumeratorCancellation] CancellationToken token = default)
+        public async IAsyncEnumerable<(byte[], int, long, long)> GetAsyncEnumerable([EnumeratorCancellation] CancellationToken token = default)
```
tpajurek-dtml, line 103: Breaking change.

cs/src/core/Index/FasterLog/FasterLogIterator.cs
```diff
@@ -97,43 +100,45 @@ internal unsafe FasterLogScanIterator(FasterLog fasterLog, BlittableAllocator<Em
...
+                yield return (result, length, currentAddress, nextAddress);
             }
         }
 
         /// <summary>
         /// Async enumerable for iterator (memory pool based version)
         /// </summary>
         /// <returns>Entry, entry length, entry address</returns>
-        public async IAsyncEnumerable<(IMemoryOwner<byte>, int, long)> GetAsyncEnumerable(MemoryPool<byte> pool, [EnumeratorCancellation] CancellationToken token = default)
+        public async IAsyncEnumerable<(IMemoryOwner<byte>, int, long, long)> GetAsyncEnumerable(MemoryPool<byte> pool, [EnumeratorCancellation] CancellationToken token = default)
```
tpajurek-dtml, line 126: Breaking change.

cs/src/core/Device/Devices.cs
badrishc: Yes, it is breaking, but the benefit is that users do not create huge files when they use the simplest API. They opt in to the preallocation. Your example did that, and it seemed wasteful, so we thought it might be better to keep preallocation as opt-in.

cs/src/core/Index/FasterLog/FasterLogIterator.cs
badrishc: This was missed out of the original API, its needed for correct async iteration with truncation.

cs/src/core/Index/FasterLog/FasterLogIterator.cs
badrishc: Same reason as above.

cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -826,29 +826,34 @@ public bool ShiftReadOnlyAddress(long newReadOnlyAddress)
...
+            // Wait for flush to complete
+            while (FlushedUntilAddress < newBeginAddress) ;
+
+            // Then shift head address
             var h = Utility.MonotonicUpdate(ref HeadAddress, newBeginAddress, out long old);
 
-            // Finally the read-only address
-            var r = Utility.MonotonicUpdate(ref ReadOnlyAddress, newBeginAddress, out old);
+            // Ensure no read cache
+            Debug.Assert(!ReadCache);
```
marius-klimantavicius, line 846: Was ReadCache always disallowed when shifting address? Maybe stronger restriction that `Debug.Assert`?

cs/src/core/Index/FasterLog/FasterLogIterator.cs
```diff
@@ -97,43 +100,45 @@ internal unsafe FasterLogScanIterator(FasterLog fasterLog, BlittableAllocator<Em
         /// Async enumerable for iterator
         /// </summary>
         /// <returns>Entry, entry length, entry address</returns>
```
hiteshmadan, line 102: Please add to comments what the fourth value in the tuple represents.

Also, I believe it is possible to add tuple member names right here, like so:

public IAsyncEnumerable<(byte[] aa, int bb, long cc, long dd)> GetAsyncEnumerable([EnumeratorCancellation] CancellationToken token = default)

cs/src/core/Index/FasterLog/FasterLogSettings.cs
```diff
@@ -77,6 +77,11 @@ public class FasterLogSettings
         /// </summary>
         public LogChecksumType LogChecksum = LogChecksumType.None;
 
+        /// <summary>
+        /// Fraction of log marked as mutable (uncommitted)
+        /// </summary>
+        public double MutableFraction = 0;
```
hiteshmadan, line 83: does this value have any effect on when commits get auto-triggered if I'm only scanning committed entries?

cs/src/core/Allocator/AllocatorBase.cs
badrishc: Correct, this is not needed; removing.

cs/src/core/Index/FasterLog/FasterLogIterator.cs
badrishc: Updating as per comment.

cs/src/core/Index/FasterLog/FasterLogSettings.cs
badrishc: The default `MutableFraction` of 0 means that every page will auto-commit as soon as it is filled up. If you have 8 pages in memory and `MutableFraction` is set to 0.5, the last 4 pages will stay mutable (uncommitted) in the absence of explicit `Commit` requests. In other words, the pages in the immutable fraction of memory are auto-committed.

Issue #296, opened 2020-08-10
Larger State Machine unit tests


tli2:
Some of the less common races within the state machine/checkpointing code wasn't caught by the current tests. Running tests with significantly more threads/clients than the number of cores on a machine will force swapping behavior and hit new interleavings that will catch more subtle races.

Issue #295, opened 2020-08-10
State Machine race for async clients


tli2:
When an async client resumes, it first reads the current state and the state machine object and fast-forwards to the current cycle. This is buggy if another thread simultaneously start a new state machine --- the async thread will capture the new state machine, but believe it to be in the old cycle, because the state it sees is REST, and it cannot differentiate between REST being the end of the last cycle and the beginning of the new cycle. Depending on the state machine being swapped in, the failure can lead to crashes or non-termination of the state machine logic.

The fix should involve capturing the state machine object in session context before suspending a thread.

badrishc:
Related latency (not correctness) issue:  a thread during `OnThreadState` in the state machine accesses global FASTER-level information such as `faster._hybridLogCheckpoint`. If the thread were a catch-up thread, this information may in fact pertain to the future, such as a future `faster._hybridLogCheckpoint.info.finalLogicalAddress`. This would cause the late thread to spin while catching up, until the ongoing (future) condition is satisfied.

PR #299, opened 2020-08-20
[C#] Checkpoint and benchmark improvements + tests

badrishc:
* Update FASTER.benchmark with more options for frequent checkpoints
* Optimize checkpointing time by adding PreallocateLog option to LogSettings
* Improve sync/async code paths in checkpoint state machine
* Fixed bugs in checkpointing logic when fast-forwarding stale client sessions
* Make checkpointing state machine sync-only, corralling async loop to `CompleteCheckpointAsync`
* Added new state machine test suite to verify state machine behavior

Fix https://github.com/microsoft/FASTER/issues/295
Fix https://github.com/microsoft/FASTER/issues/296


cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -345,9 +346,18 @@ public async ValueTask CompleteCheckpointAsync(CancellationToken token = default
                     systemState.phase == Phase.IN_PROGRESS_GROW)
                     return;
 
-                // TODO: Do we need access to functions here?
-                // If yes then move this to either faster legacy or client session.
-                await ThreadStateMachineStep<Empty, Empty, Empty, NullFasterSession>(null, NullFasterSession.Instance, true, token);
+                List<ValueTask> valueTasks = new List<ValueTask>();
+                
+                ThreadStateMachineStep<Empty, Empty, Empty, NullFasterSession>(null, NullFasterSession.Instance, ref valueTasks, token);
```
tli2, line 351: This `ref` seems unnecessary

cs/src/core/Index/FASTER/FASTERThread.cs
```diff
@@ -27,7 +27,7 @@ public partial class FasterKV<Key, Value> : FasterBase, IFasterKV<Key, Value>
                     var currentState = SystemState.Copy(ref systemState);
                     if (currentState.phase == Phase.REST)
                     {
-                        var intermediateState = SystemState.Make(Phase.INTERMEDIATE, currentState.version);
+                        var intermediateState = SystemState.Make(currentState.phase | Phase.INTERMEDIATE, currentState.version);
```
tli2, line 30: Maybe refactor this into a method for readability?

badrishc: Done

cs/src/core/Index/FASTER/FASTER.cs
badrishc: Removed the ref.

tli2:
This only partially fixes #295 , right?

badrishc:
There should no longer be any correctness issue. It's possible for a thread to wait on a future flush, but this PR ensures forward progress. A way to fix it better would be to re-check whether thread local state is in a cycle prior to the current global state, after reading the semaphore.

badrishc:
Latest check-in fixes the latency issue by preventing thread waits on future checkpoint cycles.

PR #303, opened 2020-08-28
[C++] ScanIterator issues with varlen

sillycross:
(1) The ScanIterator did not correctly handle the case that record_t is null record 
(2) The ScanIterator did not correctly handle the case that begin_address is not a multiple of page_size
(3) Compact() should skip invalid record and null record

There are more invalid assumptions (e.g. it assumes that all keys and values have fixed length), so Compact() might still be not very usable after this fix. Also the existing test for Compact() is too weak, so probably there are other hidden bugs.


badrishc:
C++ compaction and scan was written with fixed size keys and values in mind. We will consider this in the larger context of fixing compaction for varlen.

PR #305, opened 2020-08-29
[C#] Remove generic type constraint + add DataContract serializer

badrishc:
* We are removing the generic type constraint `new()` on Key and Value types in FASTER. During `Deserialize`, the user is expected to create an instance of the type being deserialized. The change is breaking for custom serializers, but is worth the benefit of relaxing type constraints.
* Changed `IObjectSerializer` `ref` params to `in` and `out` for `Serialize` and `Deserialize` respectively (breaking change for custom serializers).
* Added default serializer for generic keys and values, based on `DataContractSerializer`, in case the user does not implement `IObjectSerializer<T>`.
* Added specialized fast binary serializers for `byte[]` and `string` key/value types.
* Added tests for strings and byte arrays as key/value types.

This PR makes code like the below (specifically, we are using C# string keys directly here) work out of the box:

```cs
        static void Main()
        {
            // Create device where log will spill to disk
            var device = Devices.CreateLogDevice("foo");

            // Create simplest store instance
            using var store
                = new FasterKV<string, int>(
                    1L << 20, // size of hash table in #cache lines; 64 bytes per cache line
                    new LogSettings { LogDevice = device } // log device
                    );

            // Create new "session" to talk to store using default provided callback functions
            using var s = store.NewSession(new SimpleFunctions<string, int>());

            int key = "mykey", val = 25, inp = 0, outp = 0;

            // Synchronous upsert of value
            s.Upsert(ref key, ref val, Empty.Default, 0);
            
            // Synchronous read of value
            var status = s.Read(ref key, ref inp, ref outp, Empty.Default, 0);
            
            // Check result
            if (status != Status.OK || outp != val)
            {
                throw new Exception();
            }
        }
```

### DIFF
--- a/cs/playground/ClassCache/Types.cs
+++ b/cs/playground/ClassCache/Types.cs
@@ -34,12 +34,12 @@ namespace ClassCache
 
     public class CacheKeySerializer : BinaryObjectSerializer<CacheKey>
     {
-        public override void Deserialize(ref CacheKey obj)
+        public override void Deserialize(out CacheKey obj)
         {
             obj = new CacheKey(reader.ReadInt64());
         }
 
-        public override void Serialize(ref CacheKey obj)
+        public override void Serialize(in CacheKey obj)
         {
             writer.Write(obj.key);
         }
@@ -59,12 +59,12 @@ namespace ClassCache
 
     public class CacheValueSerializer : BinaryObjectSerializer<CacheValue>
     {
-        public override void Deserialize(ref CacheValue obj)
+        public override void Deserialize(out CacheValue obj)
         {
             obj = new CacheValue(reader.ReadInt64());
         }
 
-        public override void Serialize(ref CacheValue obj)
+        public override void Serialize(in CacheValue obj)
         {
             writer.Write(obj.value);
         }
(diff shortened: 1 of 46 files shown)


cs/playground/ClassCache/Types.cs
```diff
@@ -36,7 +36,7 @@ public class CacheKeySerializer : BinaryObjectSerializer<CacheKey>
     {
         public override void Deserialize(ref CacheKey obj)
```
marius-klimantavicius, line 37: Maybe also change `ref` to `out` to strongly indicate that deserialize needs to provide a value.

badrishc: Good idea, yes that makes sense.

badrishc: We could also symmetrically make `Serialize` take an `in` instead of `ref` object param to signify read-only input. Not sure if it might make it troublesome for some implementations of Serialize, though.  Thoughts?

badrishc: I have updated the PR to use `in` and `out` instead of `ref` for `Serialize` and `Deserialize` respectively.

PR #317, opened 2020-09-05
[C#] Add documentation of publish-subscribe with uncommitted entries

tpajurek-dtml:
Documentation related to  #255 (implemented in #298).

@badrishc


badrishc:
Looks good, thanks! I have edited and merged your write-up with what we had in the earlier section.

PR #337, opened 2020-09-22
IFunction context

aleksey-salow:
Two changes:

1. Added "ref Context" parameter to all callback functions
2. Added "ref" to Context parameter in CompletionCallback functions.

### DIFF
--- a/cs/benchmark/Functions.cs
+++ b/cs/benchmark/Functions.cs
@@ -36,26 +36,26 @@ namespace FASTER.benchmark
 
         // Read functions
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void SingleReader(ref Key key, ref Input input, ref Value value, ref Output dst)
+        public void SingleReader(ref Key key, ref Input input, ref Value value, ref Output dst, ref Empty ctx)
         {
             dst.value = value;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void ConcurrentReader(ref Key key, ref Input input, ref Value value, ref Output dst)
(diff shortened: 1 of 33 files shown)


marius-klimantavicius:
To reduce the breakage consider adding additional overloads to `ClientSession` without `ref context` where a `default(Context)` is passed. Though generally I'd wary of introducing breaking changes (I am guilty of that myself).

I guess that this is for #338.

marius-klimantavicius:
What is rationale with passing context by reference (`ref context`). If the type is class then nothing is gained, if the type is value type, then it is possible that you are going to get a reference to a copy (from pending context). 

As per https://github.com/microsoft/FASTER/issues/338#issuecomment-696846098 you could use `Input` instead of adding `Context` to all functions even though having some kind of context seem useful for other purposes than `TryAdd`.

aleksey-salow:
I'm agree. Using `Input` is much better approach than changing a lot of code.

badrishc:
I wonder if it's ever useful to add `Output` support to RMW operations, to let users copy out values after an RMW operation, to be read by user. Similar to Read. Need to see if there are good use cases for such functionality. By default, users can just do a separate Read to get the lastest value.

marius-klimantavicius:
I think it makes sense to add `Output` to RMW. I can see two scenarios - serialization is used and is expensive or we simply want to save one call (e.g. my variable length implementation does require serialization/deserialization, I am not using RMW here though); if we want to return "original" value, not the newest (consider the following - a refcount is stored in faster; if the RMWer is the first one to {interlocked} increment count, then it does some more work, otherwise it does nothing).

Issue #338, opened 2020-09-22
TryAdd equivalent


aleksey-salow:
I want implement TryAdd equivalent for Faster. After some research I found several solutions, but need advise about correct and reliable way.

1. Operation status, returned by `Session.RMW()`. If it `NOTFOUND`, raise NewRecord flag. (status is not accessible in RMWAsync)
2. `InitialUpdater` method. If method was called, than we have new record. Return flag via ???
  a. Context. Need to be added to callback functions.
  b. Output parameter, like in Read method. If we read something, we can preprocess and return it.

Custom functions for this task for RMW method:
```
public void InitialUpdater (ref Key key, ref Value input, ref Value value)
{
    value = input ;
}

public void CopyUpdater (ref Key key, ref Value input, ref Value oldValue, ref Value newValue)
{
    newValue = oldValue ;
}

public bool InPlaceUpdater (ref Key key, ref Value input, ref Value value)
{
    return false ;
}
```

Thank you.

marius-klimantavicius:
For your custom functions - `InPlaceUpdater` should return `true`, `false` is used to force a copy of record `CopyUpdater` (mostly useful for variable length structs where new value does not fix the existing record).

What you mean by 'raise NewRecord flag'?

As I understand `Input` is intended to be used as an argument to affect RMW (it does not need to be the same type as `Value`). You could use `Input` to communicate between caller and functions (be careful with value types, if data is loaded from disk you get a copy). Your input could be a `class` containing a `Value` and some kind of status. The status would be set in custom functions - for `InitialUpdater` it would indicate that a new entry was created; `CopyUpdater` and `InPlaceUpdater` would indicate that value already exists.

There is one potential inefficiency with using RMW for this - if `CopyUpdater` is used then a record is duplicated in the log. This could be undesirable, though on the other hand it also acts as a read cache of sorts as it ensures that a record is in memory (that's why it is potential inefficiency).

aleksey-salow:
> What you mean by 'raise NewRecord flag'?

Store it somewhere, so after returning from RMW it can be read. And I'm agree that `Input` can be used for this purpose.

> if `CopyUpdater` is used then a record is duplicated in the log

As I understand it can be avoided by introducing and calling new callback at next locations, before `BlockAllocate`. Making operation cancellable, if creating new record or updating existing is not desired behavior.

[...]

PR #339, opened 2020-09-24
[C#] Async version of RMW that returns status

badrishc:
* Async version of RMW that returns status
* Improved async Read to share code with sync path

Using async RMW:

Version 1:
```cs
var status = (await session.RMWAsync(ref key, ref input)).CompleteRMW();
```

Version 2:
```cs
var r = await session.RMWAsync(ref key, ref input);
while (r.status == Status.PENDING)
   r = await r.CompleteRMWAsync();
```

Version 2 is truly async even in the rare case that RMW completion has to go pending again. This is exceedingly rare (happens only when some other operation on the same key increments the value AND that value goes back to disk before this RMW could complete its `CompleteRMW`). Therefore we offer a easier version 1 just does this rare pending work synchronously, in order to keep a much simpler API.

Related issues and PRs:
* https://github.com/microsoft/FASTER/issues/338
* https://github.com/microsoft/FASTER/pull/248
* https://github.com/microsoft/FASTER/issues/246

### DIFF
--- a/cs/samples/StoreAsyncApi/Program.cs
+++ b/cs/samples/StoreAsyncApi/Program.cs
@@ -82,7 +82,7 @@ namespace StoreAsyncApi
 
                         var key = new CacheKey(rand.Next());
                         var value = new CacheValue(rand.Next());
-                        await session.UpsertAsync(ref key, ref value, context);
+                        session.Upsert(ref key, ref value, context);
                         await session.WaitForCommitAsync();
 
                         Interlocked.Increment(ref numOps);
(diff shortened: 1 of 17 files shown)


badrishc:
Proposed breaking change: rename `CompleteRead()` and `CompleteRMW()` to `Complete()` to be consistent across operations.

OLD:
```cs
(await s.ReadAsync(...)).CompleteRead();
(await s.RMWAsync(...)).CompleteRMW();
```

NEW:
```cs
(await s.ReadAsync(...)).Complete();
(await s.RMWAsync(...)).Complete();
```

badrishc:
We have removed `UpsertAsync` and `DeleteAsync` from the public API. These are not needed as these operations never go pending in the default session mode of non-thread-affinitized (`Relaxed CPR`). In thread-affinitized sessions, they may go pending in rare cases, but thread affinitized sessions cannot use C# async operations anyway, by definition.

PR #340, opened 2020-09-27
Fixed task completion source to run continuations asynchronously.

marius-klimantavicius:
Create `TaskCompletionSource` with `TaskCreationOptions.RunContinuationsAsynchronously` (all other places already do that). I noticed the error when I tried to run a highly async code (hundreds/thousands of tasks reading from faster) and hit a "deadlock", all (> 30) threads were in 
```
                while (device.Throttle())
                {
                    Thread.Yield();
                    epoch.ProtectAndDrain();
                }
```
My guess is that thread pool was exhausted thus IO completion callbacks for reading from disk were unable to be scheduled. This happened because call to `TaskCompletionSource.TrySetResult` might try to execute the continuations in the same thread (in the thread pool) and `ReadAsync` indirectly leads to the code above. As we are exclusively using the thread we might hit a point where all thread pool threads are "unavailable" for IO completion callbacks, so we never decrement `numPending`.

### DIFF
--- a/cs/src/core/Index/FASTER/FASTERImpl.cs
+++ b/cs/src/core/Index/FASTER/FASTERImpl.cs
@@ -1558,7 +1558,7 @@ namespace FASTER.core
                 request.logicalAddress = pendingContext.logicalAddress;
                 request.record = default;
                 if (asyncOp)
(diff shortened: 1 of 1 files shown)


badrishc:
The change makes sense, thanks!

PR #345, opened 2020-09-29
[C#] Added NeedCopyUpdate to IFunctions for RMW

badrishc:
The PR adds a `NeedCopyUpdate` callback to `IFunctions` for RMW, which the application can use to determine if they want to proceed with a log allocation + `CopyUpdater`. The default for this callback is `true`.

In order to force cascading into `NeedCopyUpdate`, application can just return `false` for `InPlaceUpdater`.

This will avoid unnecessary allocation if we do not need to `CopyUpdate`. The decision is made based on old value and input.

Example use cases:
* Implementing `TryAdd` functionality using RMW
* Implementing a `key->set|hash` structure (similar to redis set or even hash). We want to be able to prevent a copy if the item being added to a set is already present, otherwise if it is missing then we need to modify the set, so a copy is required. 

See linked issue below for more information.

Fix https://github.com/microsoft/FASTER/issues/338

### DIFF
--- a/cs/samples/HelloWorld/Program.cs
+++ b/cs/samples/HelloWorld/Program.cs
@@ -3,6 +3,7 @@
 
 using FASTER.core;
 using System;
+using System.Data;
 using System.IO;
 
 namespace HelloWorld
@@ -71,6 +72,26 @@ namespace HelloWorld
 
             // End session
             session.Dispose();
+
+            // (4) Perform TryAdd using RMW
+            using (var tryAddSession = store.NewSession(new TryAddFunctions<long, long>()))
+            {
+                key = 3; input1 = 30; input2 = 31;
+
+                // First TryAdd - success; status should be NOTFOUND (does not already exist)
+                status = tryAddSession.RMW(ref key, ref input1);
+                
+                // Second TryAdd - failure; status should be OK (already exists)
+                var status2 = tryAddSession.RMW(ref key, ref input2);
+
+                // Read, result should be input1 (first TryAdd)
+                var status3 = session.Read(ref key, ref output);
+
+                if (status == Status.NOTFOUND && status2 == Status.OK && status3 == Status.OK && output == input1)
+                    Console.WriteLine("(4) Success!");
+                else
+                    Console.WriteLine("(3) Error!");
+            }
 
             // Dispose store
             store.Dispose();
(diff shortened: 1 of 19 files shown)


marius-klimantavicius:
I saw that there was an attempt to implement lazy copy-updater, wondering why it failed?

badrishc:
Downsides of `LazyRef<T>` solution:

* The allocator may need to refresh epoch inside the `ref T Get()` call (due to `BlockAllocate`), which might change the world view. E.g., the `ref oldValue` might no longer be valid after an epoch refresh. This is the "critical" issue with this approach.
* If we look at the code with the split functions, it didn't seem there would be significant overlap. For instance, user can store away context in `ref Input` to avoid re-computation.
* The `LazyRef<T>` requires a number of context variables to be squirreled in before calling user code, which adds cost. E.g., the record size, execution context, FasterSession, FASTER/hlog instance. We also need to squirrel away the logical or physical address, so users can safely call `Get()` multiple times. We would need to compute/store these before calling `CopyUpdater`, even if user ends up not needing them.
* This would have been the first time we provide a reverse tunnel from user callback to FASTER - it is not clear if we want to start allowing this pattern.

Would love to hear your thoughts on this. This work isn't finalized yet, so it's worth discussing or prototyping other options to achieve the goal as well.

marius-klimantavicius:
Looking into downsides additional method is the way to go (especially due to 1, did not know that).

badrishc:
Sample end-to-end working sample use of this feature to implement `TryAdd` functionality using the in-built `TryAddFunctions`:

```cs
using var store = new FasterKV<long, long>(1L << 20, new LogSettings { LogDevice = new NullDevice() });
using var tryAddSession = store.NewSession(new TryAddFunctions<long, long>());

long key = 3, input1 = 30, input2 = 31, output = 0;

// First TryAdd - success; status should be NOTFOUND (does not already exist)
var status1 = tryAddSession.RMW(ref key, ref input1);

// Second TryAdd - failure; status should be OK (already exists)
var status2 = tryAddSession.RMW(ref key, ref input2);

// Read, result should be input1 (first TryAdd)
var status3 = tryAddSession.Read(ref key, ref output);

if (status1 == Status.NOTFOUND && status2 == Status.OK && status3 == Status.OK && output == input1)
    Console.WriteLine("Success!");
else
    Console.WriteLine("Error!");
```

PR #347, opened 2020-10-06
[C#] Extension to allow Read of all versions of key in FasterKV log

TedHartMS:
- Add IAdvancedFunctions, which adds parameters to IFunctions methods:
  - ReadCompletionCallback has an additional RecordInfo parameter
  - Reader/Writer methods have a logicalAddress parameter
- Add AdvancedClientSession, which is a parallel (not inheriting) class to ClientSession that takes an IAdvancedFunctions instance
- Add AdvancedFunctionsBase and AdvancedSimpleFunctions
- Add a DEBUG-only IClientSession interface to ensure the Session classes are consistent
- Add overloads of Read() and ReadAsync() that take a starting address that is used instead of a hash table lookup (traceback for key match is still done), and returns a copy of a found record's RecordInfo whose PreviousAddress can be passed to the next iteration of Read() to trace back a key's records. This requires IAdvancedFunctions and AdvancedClientSession
- Add ReadAtAddress and ReadAtAddressAsync to read from an address when there is no key available (the key is passed to the IFunctions or IAdvancedFunctions Read methods)


cs/samples/ReadAddress/ReadAddressApp.cs
```diff
@@ -0,0 +1,206 @@
...
+            sw.Start();
+            var prevLap = 0;
+            for (int ii = 0; ii < numKeys; ii++)
+            {
+                // lap is used to illustrate the changing values
+                var lap = ii / keyMod;
+
+                if (useCheckpoints && lap != prevLap)
+                {
+                    store.TakeFullCheckpoint(out _, CheckpointType.FoldOver);
```
badrishc, line 123: Use `await store.TakeFullCheckpointAsync` so that this checkpoint is completed before the next one is initiated.

Issue #329, opened 2020-09-14
[C#] FasterKV for Span<byte> keys and values


AlgorithmsAreCool:
Do you think there would be value in a primitive like a `FasterKV` but that only operated on `Span<byte>` for the keys and values?

I'm migrating some test code from LMDB and really came to appreciate the simplicity of just working with blobs.

badrishc:
Yes, absolutely. This has been on our radar for a while. It's actually not hard to pull off as a layer over our 'varlen' struct support.

Related, FasterLog already works on top of Span<byte>, see here: https://github.com/microsoft/FASTER/blob/master/docs/cs/FasterLog.md

mito-csod:
In the same spirit, what do you think about supporting Memory<T> and ReadOnlyMemory<T> in FasterLog.EnqueueAsync(),, @badrishc? It would let us use pooled memory chunks we might have gotten elsewhere without the extra copying to byte[], and still let us use async. Happy to send a PR if you think it's a worthwhile addition.

badrishc:
> In the same spirit, what do you think about supporting Memory and ReadOnlyMemory in FasterLog.EnqueueAsync(),, @badrishc? It would let us use pooled memory chunks we might have gotten elsewhere without the extra copying to byte[], and still let us use async. Happy to send a PR if you think it's a worthwhile addition.

Yes, that would definitely be a worthwhile addition!

mito-csod:
My bad, it looks like the ReadOnlyMemory implementation is already there--intellisense tricked me!

PR #349, opened 2020-10-10
[C#] SpanByte - a safe wrapper for variable-length keys and values

badrishc:
We introduce two types in FASTER that make it easy to store and retrieve variable-length keys and values with zero-copy, no serializer, and with in-place allocation on a single Faster record log (no object log)

## SpanByte

A new variable-length struct type called `SpanByte` has been added to core FASTER, in order to make it easy to safely use variable-length keys and values in FasterKV.

`SpanByte` represents a variable length byte payload in two data formats:
* In stack mode, `SpanByte` represents a combination of `length` and a pointer to fixed data
* In serialized mode, `ref SpanByte` represents a pointer to serialized structure [length, payload]

Internally, we use one bit in the common length header to distinguish between the two formats.

We use `SpanByte` instead of `Span<byte>` and `Memory<byte>` for FasterKV key and value types because:
* `Span<byte>` cannot be used as a generic type
* we expect the same types to be serialized to disk in-place on the log
* `Memory<byte>` cannot point to fixed stack or unmanaged memory buffers, which are necessary for zero-copy

It is extremely easy to convert any fixed `Span` or pinned `Memory` to `SpanByte` (we provide converters), as our sample shows [here](https://github.com/microsoft/FASTER/tree/spanbyte/cs/samples/StoreVarLenTypes). We also provide extension methods to `ClientSession` to accept `Span<byte>` directly and perform the pinning internally.

## SpanByteAndMemory

We also provide a type called `SpanByteAndMemory` that represents a union of `SpanByte` and `IMemory<byte>`. This type can be used as your `Output` type in FasterKV. The reason for this type is to enable users to have a fixed region of memory (usually in stack) and perform a `Read` operation that can directly write to that location if space permits. If no space, or the operation goes async (due to I/O), we will instead return the output via `IMemory<byte>` allocated from a shared pool.

<br/>
Note, the underlying technology is the variable-length struct support already available in FASTER. The PR extends varlen to support a log serialization format for varlen types, that is possibly different from the user's data format. `SpanByte` and `SpanByteAndMemory` make it easier to use varlen out of the box, with fast and safe application code that can leverage the `Span` and `Memory` features of C#.

Fix https://github.com/microsoft/FASTER/issues/329


badrishc:
Yes, `Span<byte>` (and therefore, `Memory<byte>`) is supported via extension methods. You create a `FasterKV<SpanByte, SpanByte>`, these these extension methods on session become available to you. See [this example](https://github.com/microsoft/FASTER/blob/2c699a3f4fa2ab70c56d25bd3fa6f13692fec926/cs/samples/StoreVarLenTypes/Program.cs#L64). The PR description explains why `Memory<byte>` cannot directly be used as our FASTER key/value type (related to needing a disk serialized version of the same type).

badrishc:
Oh, I also think you can use `Memory<byte>` directly with FasterKV if you are ok with that. We will provide an example shortly.

AlgorithmsAreCool:
Yes, I think your design is spot on. Just noting a tiny difference in the statement:

> Memory<byte> cannot point to fixed stack or unmanaged memory buffers, which are necessary for zero-copy

Sorry for introducing confusion 😅

This is great stuff!

badrishc:
Agreed, `Memory` can indeed point to unmanaged buffers. In fact, our "managed storage device" uses this facility to call .NET storage API using Memory<T> to write unmanaged pages to disk.

badrishc:
Great suggestions, @AlgorithmsAreCool. We are adding support in this PR: https://github.com/microsoft/FASTER/pull/354

It may not be as hyper-efficient as `SpanByte`, but will be much easier and robust to use, as `Memory<byte>` is easily understood by C# developers. This way you get a "proper" variable-length key-value store API.

PR #354, opened 2020-10-17
[C#] Support Memory<T> keys and values out of the box

badrishc:
Easy-to-use variable-length data support in FASTER with a single backing hybrid log (no object log). Adding support for `Memory<T> where T : unmanaged` as value type in FasterKV. You can use `ReadOnlyMemory<T>` or `Memory<T>` as key type. You can write code like this:

```cs
var store = new FasterKV<ReadOnlyMemory<byte>, Memory<byte>>(...);
var s = store.For(new MyMemoryFunctions()).NewSession<MyMemoryFunctions>();
var key = new ReadOnlyMemory<byte>(new byte[25]);
key.Span.Fill(1);
var value = new Memory<byte>(new byte[50]);
value.Span.Fill(2);
s.Upsert(key, value);
```

The default `MemoryFunctions` implementation accepts a `MemoryPool<T>` and uses `(IMemoryOwner<T> memoryOwner, int length)` as its `Output` type. This allows you to use a shared memory pool to store the returned values.


cs/src/core/Index/FASTER/FASTERImpl.cs
```diff
@@ -1401,34 +1400,35 @@ internal enum LatchOperation : byte
...
+                    (actualSize, allocatedSize) = hlog.GetInitialRecordSize(ref key, ref pendingContext.input.Get(), fasterSession);
                 }
                 else
                 {
                     physicalAddress = (long)request.record.GetValidPointer();
-                    recordSize = hlog.GetRecordSize(physicalAddress, ref pendingContext.input.Get(), fasterSession);
+                    (actualSize, allocatedSize) = hlog.GetRecordSize(physicalAddress, ref pendingContext.input.Get(), fasterSession);
                 }
-                BlockAllocate(recordSize, out long newLogicalAddress, sessionCtx, fasterSession);
+                BlockAllocate(actualSize, out long newLogicalAddress, sessionCtx, fasterSession);
```
marius-klimantavicius, line 1413: Shouldn't this be `allocatedSize` (it seems to be aligned to record size, thus possibly bigger than `actualSize`, I haven't checked does `BlockAllocate` align the size to record)?

badrishc: Great catch, yes!

badrishc:
Plan to generalize this to work with `Memory<T> where T  : unmanaged`, in the next check-in. The nice thing is that `IVariableLengthStruct` can now work with any type as long as it has a log-serialized equivalent.

cs/src/core/Index/FASTER/FASTERImpl.cs
badrishc: We need to allocate to 8-byte-aligned size so that log full scan can reasonably skip "zero byte" segments of the log in 8-byte increments.

marius-klimantavicius:
When I was thinking about `Memory<T>` I thought that it makes sense to have a separate allocator which actually prepends length and the copies the content from buffer (this did not go anywhere but my reasoning was that I wanted to be able to use buffer without worrying about length prefix and prevent additional copies as much as I could).

badrishc:
> When I was thinking about `Memory<T>` I thought that it makes sense to have a separate allocator which actually prepends length and the copies the content from buffer (this did not go anywhere but my reasoning was that I wanted to be able to use buffer without worrying about length prefix and prevent additional copies as much as I could).

This PR does exactly that, using existing varlen allocator. We allocate `memory.Length + sizeof(int)` on the log. Then, the user callbacks create a temporary `Memory<byte>` whose payload points directly to the value pointer + sizeof(int) offset.

badrishc:
FYI, `IVariableLengthStruct<T>` has a couple of new calls added to it. These calls are currently based on unsafe `void*` pointers. They should eventually move to use `Span<byte>` based API for safety and simplicity for implementers of this interface.

Current:
```cs
unsafe void Serialize(ref T source, void* destination)
unsafe ref T AsRef(void* source)
unsafe void Initialize(void* source, void* end)
```

Should move to:
```cs
void Serialize(ref T source, Span<byte> destination)
ref T AsRef(Span<byte> source)
void Initialize(Span<byte> source)
```

badrishc:
Added locking capability to values for both `SpanByte` and `Memory<T>`, using bits in the length header. You can also mark a record as read-only (under lock) during a concurrent update or IPU, so that the next operation will be forced to be an allocation on tail. For example, we set the record as read-only when `InPlaceUpdater` or `ConcurrentReader` is about to return false. See the example for details: https://github.com/microsoft/FASTER/tree/memory-support/cs/samples/StoreVarLenTypes

Issue #358, opened 2020-10-24
(1.7.2) NotImplementedException thrown sometimes in VarLenBlittableAllocator.cs when deleting entry, using SpanByte as key and value


resetcoder:
I'm using `FasterKV<SpanByte, SpanByte>` the newest release (1.7.2), having a fully functional project, but sometimes I get `System.NotImplementedException` in [VarLenBlittableAllocator.cs:line 191](https://github.com/microsoft/FASTER/blob/master/cs/src/core/Allocator/VarLenBlittableAllocator.cs#L191) when deleting entries in bulk.

When I'm deleting them one by one manually everything works fine, but if I iterate through keys from a List<> or Dictionary<>, or just calling it many times during stress-tests I always get this exception.

**Stacktrace:**
```
System.NotImplementedException
  HResult=0x80004001
  Message=The method or operation is not implemented.
  Source=FASTER_Test2
  StackTrace:
   at FASTER.core.VariableLengthBlittableAllocator`2.ShallowCopy(Key& src, Key& dst) in ...\FASTER\core\Allocator\VarLenBlittableAllocator.cs:line 191
   at FASTER.core.FasterKV`2.InternalDelete[Input,Output,Context,FasterSession](Key& key, Context& userContext, PendingContext`3& pendingContext, FasterSession fasterSession, FasterExecutionContext`3 sessionCtx, Int64 lsn) in ...\FASTER\core\Index\FASTER\FASTERImpl.cs:line 1029
   at FASTER.core.FasterKV`2.ContextDelete[Input,Output,Context,FasterSession](Key& key, Context context, FasterSession fasterSession, Int64 serialNo, FasterExecutionContext`3 sessionCtx) in ...\FASTER\core\Index\FASTER\FASTER.cs:line 537
   at FASTER.core.ClientSession`6.Delete(Key& key, Context userContext, Int64 serialNo) in ...\FASTER\core\ClientSession\ClientSession.cs:line 415
   at KV.KVStore.WorkThread.Loop() in ...\FASTER\KVStore.cs:line 2370
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state) in /_/src/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs:line 50
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state) in /_/src/System.Private.CoreLib/shared/System/Threading/ExecutionContext.cs:line 172
[...]

PR #359, opened 2020-10-24
[C#] Compaction via session, update EntryCount, remove ShallowCopy

badrishc:
* Move `Compact` to `ClientSession` so that user does not need to implement Copy/InPlaceCopy functions. Example:
  ```cs
  using (var session = store.NewSession(...))
    session.Compact(address, true);
  ```
* Update `EntryCount` and `DumpDistribution` to reflect only hash entries >= `Log.BeginAddress`
* (Internal) Remove ShallowCopy since we now have serialize

Fix https://github.com/microsoft/FASTER/issues/358


badrishc:
This PR also simplifies log compaction. Basically, it moves the `Compact` call to `ClientSession`. This way, we can use the existing `IFunctions` to perform the compaction. These functions would already handle concurrency for normal operations. What this also implies is that `ICompactionFunctions` can be simplified to only contain `IsDeleted(...)` and user does not need to provide anything by default.

Usage Example

```cs
using (var session = store.NewSession(...))
   session.Compact(address, true);
```

We can eventually deprecate the `Compact` operation on `LogAccessor` (i.e., `store.Log.Compact(...)`) - when we remove it, we can also remove the `Copy` and `InPlaceCopy` methods from the `ICompactionFunctions` interface.

Issue #231, opened 2020-01-10
[C++] Support io_uring for Linux IO


badrishc:
In the current version of FASTER C++, on Linux, we use `libaio` for async IO handling. It is a known issue that `libaio` is not very efficient. Recently, `io_uring` is released with Linux kernel 5.1, which advertises to be a high performance aysnc IO library. It would be useful to try it and see if we can improve disk performance on Linux by replacing `libaio` with it.

More details for io_uring can be found here: http://kernel.dk/io_uring.pdf

dongx-psu:
I think I got some time recently on this. Will try get a version out this month.

PR #387, opened 2020-12-11
[C++] Linux liburing based I/O handler.

dongx-psu:
Add a liburing based I/O handler for C++ version of FASTER. Need kernel version > 5.1.

Add two corresponding unit test case as well.

Fix https://github.com/microsoft/FASTER/issues/231


dongx-psu:
This may need some benchmarking to see the perf improvement. Resolving #231

badrishc:
Thanks for the PR. We can evaluate it on the YCSB benchmark with small memory allocation, to see how the #threads vs IOPS graph looks like.

dongx-psu:
Made `USE_URING` an option (by default off) so that user can choose to build it with it or not.

PR #401, opened 2021-02-07
fix(iterator): null conditional dispose

stephankempkes:
In some occasions iter2 is null, which leads to not disposing any other members. This check only disposes members when they are not null.


badrishc:
Correct,  merged. Thanks!

PR #413, opened 2021-03-03
[C++] Faster calculation of free thread id

aabbdev:
Hi,

I changed the limit of kMaxNumThreads from 96 to 128 to allow 1 cycle operations, 40/50% faster.
And I modified the modulos in door AND.
it will allow the compiler to also add its own optimizations.
And today's processors can easily have up to 128 threads (AMD Threadripper 3390WX)

Sorry for the inconvenience ^^
I may have questions to ask you, if you have time.

### DIFF
--- a/cc/src/core/thread.h
+++ b/cc/src/core/thread.h
@@ -21,7 +21,7 @@ class Thread {
   /// The number of entries in table. Currently, this is fixed at 96 and never changes or grows.
   /// If the table runs out of entries, then the current implementation will throw a
   /// std::runtime_error.
-  static constexpr size_t kMaxNumThreads = 96;
+  static constexpr size_t kMaxNumThreads = 128; // 50% faster with a power value of 2, to be able to replace modulo by AND 
 
  private:
   /// Encapsulates a thread ID, getting a free ID from the Thread class when the thread starts, and
@@ -58,8 +58,8 @@ class Thread {
     uint32_t end = start + 2 * kMaxNumThreads;
     for(uint32_t id = start; id < end; ++id) {
       bool expected = false;
-      if(id_used_[id % kMaxNumThreads].compare_exchange_strong(expected, true)) {
-        return id % kMaxNumThreads;
+      if(id_used_[id & (kMaxNumThreads - 1)].compare_exchange_strong(expected, true)) {
+        return id & (kMaxNumThreads - 1);
       }
     }
     // Already have 64 active threads.


badrishc:
Does it affect the performance benchmark, due to the larger epoch table size?

aabbdev:
This will reduce latency, there will be less holes in the execution thread. Modulo is extremely expensive, you should do everything to avoid using this instruction if possible.
It can be replaced by the operation &, if the operand on right is a power of two. That's why I increased the constant to 128 instead of 96.

Elsewhere there are more and more processors with 128 threads.

badrishc:
Yes we are aware of this and use this in many places throughout the codebase in all critical places. In this particular location, the optimization does not help as this check occurs very infrequently. Are you noticing a performance loss, or are you running out of threads, in your use case?

aabbdev:
I was just correcting the code, from my point of view an optimization is not necessarily a gain +30% on the whole program. but 1% + 1% +1% [..] = 20%

aabbdev:
Even if this part is not often performed, when it is, it will be performed very quickly.

So far I've only compiled locally, and I was reading the code to be able to replace the thread pool with a fiber task scheduler (hybrid thread).

This would allow me to have only one thread pool in my program and not to use the kernel to manage threads but to make everything user-side (much faster).

-> https://github.com/google/marl

badrishc:
Thank you for the interesting points. Perhaps I was not clear about my concern:

This change comes with a performance trade-off. Quite frequently (once every epoch refresh), FASTER does a full scan of the epoch table in order to move the safe epoch forward. If the epoch table is larger (128 vs 96) this full scan takes longer.

The benefit of a larger epoch table is more threads (up to 128) can use FASTER.

We found this overhead could be large enough to affect system throughput in the YCSB benchmark. That's why we chose 96, since the system at the time had less than 96 threads.

Users can change this constant if they are using larger thread counts.

The question is: does going from 96 to 128 as default  give us enough benefit to compensate for the lower throughput.

badrishc:
By the way, I really like the idea of replacing the thread pool with fiber task scheduler. It will remove a key limitation of the codebase. Look forward to your findings and contributions :).

aabbdev:
I may have a solution,
use two 64-bit variables, and use each bit as an id.
If a bit == 1 then it is used.

aabbdev:
We could use SIMD to do 4 checks at once like:

load 4 bytes
output = simd_bitwise_and([i, i+1, i+2, i+3], [128, 128, 128, 128])
it will reduce the number of memory accesses by 4x and do 4x more AND per cycle.

badrishc:
Please see light_epoch.cc very carefully. Epoch protection is designed like that for a strong reason.

Each entry has to be one cache line or larger. Individual threads are updating their thread local counter very frequently in their own epoch entry. Making it smaller will make cache coherence go berserk due to the false sharing across threads. The system will no longer scale with number of threads.

Also read the FASTER paper for details: https://microsoft.github.io/FASTER/docs/td-research-papers/ and there might be older papers on epoch protection as well.

aabbdev:
Very interesting, I would watch.
Otherwise reduce to 64 threads according to the benchmark after 60 threads the scaling doesn't work well anymore.

badrishc:
We are running experiments on 72 core machines. Scaling will work well until > 256 cores, we just need to set the table size appropriately for each experiment. The point is: we should not "pay" up front for higher core count when running on a machine with fewer core count. So we chose 96 as a reasonable compromise. If you have a use case with >96, then just increase that to 128 or 250 or whatever.

badrishc:
Flattening of scaling depends on workload. For large dataset, you may be memory bandwidth limited after 64 threads using non-trivial key and value sizes. Epoch protection itself is quite scalable at least for around 200 entries.

aabbdev:
really interested thank you very much for taking the time to answer me,
I think then the implementation of fiber based threads will be interesting:
- No significant increase in memory compared to the number of threads.
- Switch between threads incredibly faster
- from 1 to a hundred cores without any problem.
- Threads have their own error stack.
 
There is shovel of advantage.

badrishc:
sounds good - look forward to your findings 👍

PR #415, opened 2021-03-04
[C++] Faster.h - Optimization power of 2

aabbdev:
We should do a benchmark, it should be faster now.

EDIT: I made these modifications from my browser I will compile locally to see

### DIFF
--- a/cc/src/core/faster.h
+++ b/cc/src/core/faster.h
@@ -58,14 +58,14 @@ class alignas(Constants::kCacheLineBytes) ThreadContext {
   }
 
   inline const ExecutionContext& prev() const {
-    return contexts_[(cur_ + 1) % 2];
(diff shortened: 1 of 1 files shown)


aabbdev:
I have applied this optimization almost everywhere in the code, I have gained 200 ms on a memory test, I just have to put the segments in power of 2

badrishc:
Note that if a value is already a power of 2, C++ compiler automatically performs the optimization to avoid %. So code readability can be maintained.

aabbdev:
I didn't finish the commit, but yes indeed the code is automatically transformed if the constant is a power of 2.
I have modified constants like segments to values close to the power of 2.

Sorry for the inconvenience ^^

badrishc:
No problem, look forward to your findings. You can fork the code, complete the work and then send us a PR to review.

PR #419, opened 2021-03-06
Quite a few new tests added for Faster Log and Faster KV to get functional coverage 

darrenge:
Added tests to make sure every feature for Faster Log and Faster KV have at least one corresponding test.


darrenge:
I see there are other changes to Master that I didn't have. Will do due diligence and merge that before doing a PR

Issue #417, opened 2021-03-04
[C#] Cannot delete fasterlog from disk if the log was disposed with a commit request in flight


hiteshmadan:
I recently upgraded to FASTER's latest nuget version and the following testcase started failing. 

The testcase initiates a write, initiates a commit (with spinWait=false), disposes off the log and the device, and then tries to delete the log from disk.  This can easily happen in my app because the app has API's to create/delete logs and to append to logs. The delete and append can be done concurrently.

The delete fails with `System.IO.IOException: The process cannot access the file 'commit.0.0' because it is being used by another process.`

I figured out that there is some code in fasterlog that's holding an active reference to the log-commit file even after dispose.

Note that if I remove the call to log.Commit, or do the commit with spinWait=true, then the delete does not fail.

Note that even if i wait for the commit to complete (with spinWait=false) - by checking that the ValueTask returned by TryEnqueue has succeeded/failed - I still cannot delete the log.

```
[TestMethod]
public void TestDisposeReleasesFileLocksWithInprogressCommit()
{
    DirectoryInfo di = Directory.CreateDirectory(this.TestContext.TestRunDirectory + $"\\log");
    IDevice device = Devices.CreateLogDevice(this.TestContext.TestRunDirectory + $"\\log\\testDisposeReleasesFileLocksWithInprogressCommit.log", preallocateFile: true, deleteOnClose: false);
    FasterLog fasterLog = new FasterLog(new FasterLogSettings { LogDevice = device, LogChecksum = LogChecksumType.PerEntry });
    fasterLog.TryEnqueue(new byte[100], out long beginAddress).Should().BeTrue();
    fasterLog.Commit(spinWait: false);
    fasterLog.Dispose();
    device.Dispose();
    di.Delete(recursive: true);
}
```

badrishc:
`Commit` initiates a write to disk, so the delete would fail until that write completes. The following seems to work:

```cs
        [Test]
        public void TestDisposeReleasesFileLocksWithInprogressCommit()
        {
            var commitPath = TestContext.CurrentContext.TestDirectory + "/" + TestContext.CurrentContext.Test.Name + "/";
            DirectoryInfo di = Directory.CreateDirectory(commitPath);

            IDevice device = Devices.CreateLogDevice(commitPath + "fasterlog.log", preallocateFile: true, deleteOnClose: false);
[...]

PR #423, opened 2021-03-08
[C#] Support concurrent device dispose and async commit

badrishc:
Fix https://github.com/microsoft/FASTER/issues/417

### DIFF
--- a/cs/src/core/Device/LocalStorageDevice.cs
+++ b/cs/src/core/Device/LocalStorageDevice.cs
@@ -37,6 +37,7 @@ namespace FASTER.core
         private readonly bool useIoCompletionPort;
         private readonly ConcurrentQueue<SimpleAsyncResult> results;
         private static uint sectorSize = 0;
+        private bool _disposed;
 
         /// <summary>
         /// Number of pending reads on device
@@ -105,6 +106,8 @@ namespace FASTER.core
 #endif
             ThrottleLimit = 120;
             this.useIoCompletionPort = useIoCompletionPort;
+            this._disposed = false;
+
             if (useIoCompletionPort)
             {
                 ThreadPool.GetMaxThreads(out int workerThreads, out _);
@@ -329,6 +332,7 @@ namespace FASTER.core
         /// </summary>
         public override void Dispose()
         {
+            _disposed = true;
             foreach (var logHandle in logHandles.Values)
                 logHandle.Dispose();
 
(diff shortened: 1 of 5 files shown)


cs/test/FasterLogTests.cs
```diff
@@ -13,6 +13,33 @@
...
+            fasterLog.Dispose();
+            device.Dispose();
+            while (true)
+            {
+                try
+                {
+                    di.Delete(recursive: true);
+                    break;
+                }
+                catch { }
```
hiteshmadan, line 38: interesting, we still expect an exception? 
Did I get it wrong that after the current fix there'll be no exceptions in DirectoryInfo.Delete() ?

hiteshmadan (approved):

cs/test/FasterLogTests.cs
badrishc: This sequence, though rare, could happen:
1. Async IO thread calls `Commit`, which first checks `_disposed` to be false
2. `Dispose` sets `_disposed` to be true and disposes the file handle
3. `Commit` sees the handle is null, creates a new handle
4. Main thread tries to delete directory, fails
5. `Commit` thread re-checks `_dispose` and deletes the handle it just created (for handle reclamation correctness).

To avoid this case, we retry the directory deletion (it did not happen in the tests though). Note that I removed the `WaitForCommit` and the check for `IsFaulted` state in this testcase -- if you had that then this loop would not be needed.

Also, unrelated, we did seem to notice earlier (though not confirmed) that the OS releases the handle a bit later after the dispose is called, i.e. an immediate directory deletion does not always work. This is unclear though.

cs/src/core/Device/LocalStorageDevice.cs
```diff
@@ -445,7 +449,15 @@ protected SafeFileHandle GetOrAddHandle(int _segmentId)
             {
                 return h;
             }
-            return logHandles.GetOrAdd(_segmentId, segmentId => CreateHandle(segmentId));
+            if (_disposed) return null;
```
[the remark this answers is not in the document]
badrishc: Correct, the call sites of `WriteFile` and `ReadFile` catch the exception when reading or writing to a null handle, and issue the IO callback on the current thread itself:

```cs
            catch
            {
                Interlocked.Decrement(ref numPending);
                callback(uint.MaxValue, 0, context);
                results.Enqueue(result);
            }
```

Issue #434, opened 2021-03-16
[C#] Allow PendingFlushList's size to be configurable/dynamic


hiteshmadan:
We see that when the storage layer is running slow + there are a lot of frequent commit() calls done by the app, the pendingFlushList tends to fill up quickly.

Two options were discussed: 
- make the size of the list configurable (its hardcoded to 8 right now)
- have an "elastic flush list" implementation that merges adjacent flushes into one to free up space

Issue #433, opened 2021-03-16
[C#] Handle FileStream flush/dispose exceptions in ManagedLocalStorageDevice


hiteshmadan:
MLSD.Read and Write do a StreamPool.Return on a filestream object, which results in the filestream being disposed.
A filestream dispose internally calls flush in some cases which can lead to exceptions esp if we're dealing with a remote file share mounted to local fs.

Any exceptions in flush/dispose will result in the DeviceIOCallback never being invoked, which per offline discussions with Badrish can be a harmful state to be in.

Opening an issue to make sure that any such exceptions are correctly handled.

PR #436, opened 2021-03-17
[C#] MLSD and FasterLog improvements

badrishc:
* Better handling of async tasks in `ManagedLocalStorageDevice` to work efficiently on Linux with large number of tasks.
* Avoid exception on full `PendingFlushList` and instead fail the commit.
* Use locks to correctly invoke `FileStream.ReadAsync` and `FileStream.WriteAsync` in MLSD.
* Merge adjacent unissued commit requests in `PendingFlushList` for improved commit behavior.
* Remove spin from `FixedPool` when no space left, and instead issue new object for immediate disposal.

Fix https://github.com/microsoft/FASTER/issues/434
Fix https://github.com/microsoft/FASTER/issues/433
Co-authored with @hiteshmadan

### DIFF
--- a/cs/src/core/Device/ManagedLocalStorageDevice.cs
+++ b/cs/src/core/Device/ManagedLocalStorageDevice.cs
@@ -215,9 +215,12 @@ namespace FASTER.core
                     await logWriteHandle.WriteAsync(umm.Memory);
 #else
                     SectorAlignedMemory memory = pool.Get((int)numBytesToWrite);
-                    fixed (void* destination = memory.buffer)
+                    unsafe
                     {
-                        Buffer.MemoryCopy((void*)sourceAddress, destination, numBytesToWrite, numBytesToWrite);
+                        fixed (void* destination = memory.buffer)
+                        {
+                            Buffer.MemoryCopy((void*)sourceAddress, destination, numBytesToWrite, numBytesToWrite);
+                        }
                     }
 
                     await logWriteHandle.WriteAsync(memory.buffer, 0, (int)numBytesToWrite);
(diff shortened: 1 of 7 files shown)


cs/src/core/Device/ManagedLocalStorageDevice.cs
```diff
@@ -97,65 +104,107 @@ private void RecoverFiles()
...
-                    callback(errorCode, (uint)t.Result, context);
+                    // Sequentialize all reads from same handle on non-windows
+                    if (!IsWindows)
+                    {
+                        streampool?.Return(offset);
+                    }
+                    streampool?.DisposeIfNeeded(offset, logReadHandle);
                 }
-                );
+                catch (Exception ex)
```
hiteshmadan, line 180: In case of any exceptions in await readTask, the streamPool.Return (when IsWindows=false) and streamPool.DisposeIfNeeded will not be invoked. Is that intentional?

cs/src/core/Device/ManagedLocalStorageDevice.cs
```diff
@@ -97,65 +104,107 @@ private void RecoverFiles()
...
+                    numBytes = 0;
+                }
+                finally
+                {
+                    Interlocked.Decrement(ref numPending);
+#if !NETSTANDARD2_1
+                    unsafe
+                    {
+                        fixed (void* source = memory.buffer)
+                            Buffer.MemoryCopy(source, (void*)destinationAddress, numBytes, numBytes);
```
hiteshmadan, line 201: Why should this memory copy happen when 'await readTask' throws an exception?

cs/src/core/Device/ManagedLocalStorageDevice.cs
badrishc: Fixed, please re-check code.

cs/src/core/Device/ManagedLocalStorageDevice.cs
badrishc: Fixed, please re-check code.

PR #443, opened 2021-03-26
[C#] Fasterlog: Remove thread.yield from FasterLog.NeedToWait()

hiteshmadan:
EnqueueAndWaitForCommitAsync() tends to make a lot of attempts at TryEnqueue (20 times per call to EWC in my runs on a 4 core alpine container with 8 GB RAM, 2 MB page size, 4 pages in memory) - doing a Thread.Yield() for every iteration of that tight loop burns CPU a lot.


badrishc:
Closing as subsumed by https://github.com/microsoft/FASTER/pull/446

PR #450, opened 2021-04-14
[C#] Add Status to CompletedOutput

TedHartMS:
Need to indicate status of the operation so the caller knows if the key was found


TedHartMS:
Closing - redo in new branch

PR #486, opened 2021-06-02
[C#] Update to C#9

TedHartMS:
- Add <LangVersion>latest to csprojs to pick up C#9
- Remove "is {}" in favor of "is not null"
- Use "Type x = new()", which is particularly useful for field definitions


TedHartMS:
Fixed Linux builds

badrishc (approved):

PR #497, opened 2021-06-10
[C#] [WIP] Pub/sub functionality in remote FasterKV

badrishc:
* Ability to subscribe to specific keys in the database
* User will get updates to keys that are updated via Upsert or RMW
* Pub/sub session is to be kept separate from normal get/put sesstions

Being co-authored by Rohan Kadekodi.


badrishc:
Closing as superceded by https://github.com/microsoft/FASTER/pull/514

PR #507, opened 2021-06-22
[C#] Parallel test fixtures

TedHartMS:
Enable Unit Test parallelization at the TestFixture level.
- For NET461, throttle LevelOfParallelism to 8 as it runs x86
- Also use consistent patterns:
  - Added TestUtils properties to get test folder and test container names, which are based on TestClass/TestMethod for uniqueness
    - This is modified to break out "FASTER.test" as the top-level directory in the test hierarchy, followed by class name (minus FASTER.test), followed by method name. This makes it easy to clean up by just removing <bindir>\FASTER.test
  - Made all tests use TestUtils.DeleteDirectory
  - Make sure Setup doesn't use prior values
    - Prefer allocations in Setup to field initializers
  - null-safety in TearDown
  - Break up a couple big test classes so the Fixtures can run in parallel (FasterLogTests, RecoveryChecks)
- Add Category("CheckpointRestore") in a few more places
- Make sure directories are cleaned up on test completion

Tests complete about 4x faster now on my machine


darrenge (commented):
I looked this over and I am ok with pushing this to Master.  Badrish ... you make the final call.

TedHartMS:
Subsumed by #523

PR #529, opened 2021-07-23
[C#] Faster Extensions for DPR Integration (WIP)

tli2:
Adds the following feature to integrate DPR into FASTER:

- Custom actions to run before entering state machine stages
- Custom commit cookie to write along with each checkpoint's commit data
- Checkpointing from v -> arbitrary version instead of just v + 1
- Handle version roll-over corner case
- Expose version numbers as public Properties

Will test functionality before merge

### DIFF
--- a/cs/src/core/Index/Common/Contexts.cs
+++ b/cs/src/core/Index/Common/Contexts.cs
@@ -405,15 +405,16 @@ namespace FASTER.core
         /// <param name="token"></param>
         /// <param name="checkpointManager"></param>
         /// <param name="deltaLog"></param>
-        /// <returns></returns>
-        internal void Recover(Guid token, ICheckpointManager checkpointManager, DeltaLog deltaLog = null)
+        /// <returns> Any user-specified commit cookie written as part of the checkpoint </returns>
+        internal byte[] Recover(Guid token, ICheckpointManager checkpointManager, DeltaLog deltaLog = null)
         {
             var metadata = checkpointManager.GetLogCheckpointMetadata(token, deltaLog);
             if (metadata == null)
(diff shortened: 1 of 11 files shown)


cs/src/core/Index/Common/Contexts.cs
```diff
@@ -428,15 +428,16 @@ public void Initialize(StreamReader reader)
         /// <param name="token"></param>
         /// <param name="checkpointManager"></param>
         /// <param name="deltaLog"></param>
-        /// <returns></returns>
-        internal void Recover(Guid token, ICheckpointManager checkpointManager, DeltaLog deltaLog = null)
+        /// <returns> Any user-specified commit cookie written as part of the checkpoint </returns>
+        internal byte[] Recover(Guid token, ICheckpointManager checkpointManager, DeltaLog deltaLog = null)
```
badrishc, line 432: Better to be an overload of recover that returns void, but includes an `out commitCookie` parameter so it is obvious what the array is.

cs/src/core/Index/CheckpointManagement/DeviceLogCommitCheckpointManager.cs
```diff
@@ -323,7 +323,7 @@ public byte[] GetLogCheckpointMetadata(Guid logToken, DeltaLog deltaLog)
             else
                 ReadInto(device, 0, out body, size + sizeof(int));
             device.Dispose();
-            return new Span<byte>(body).Slice(sizeof(int)).ToArray();
+            return body.Skip(sizeof(int)).Take(size).ToArray();
```
badrishc, line 326: Avoid Skip, use Slice(start, length) with Span.

PR #535, opened 2021-08-04
[C#] Checkpoint & Versioning Utilities

tli2:
- Add ability to recover to a specific version within an incremental snapshot
- Add ability to remove a particular checkpoint guid from ICheckpointManager instead of just PurgeAll()

### DIFF
--- a/cs/src/core/Index/Synchronization/HybridLogCheckpointTask.cs
+++ b/cs/src/core/Index/Synchronization/HybridLogCheckpointTask.cs
@@ -174,7 +174,6 @@ namespace FASTER.core
     /// </summary>
     internal sealed class SnapshotCheckpointTask : HybridLogCheckpointOrchestrationTask
     {
-        private long lastVersion;
         /// <inheritdoc />
         public override void GlobalBeforeEnteringState<Key, Value>(SystemState next, FasterKV<Key, Value> faster)
         {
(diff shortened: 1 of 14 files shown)


cs/src/core/Index/Synchronization/HybridLogCheckpointTask.cs
```diff
@@ -217,6 +221,7 @@ internal sealed class SnapshotCheckpointTask : HybridLogCheckpointOrchestrationT
                         out faster._hybridLogCheckpoint.flushedSemaphore);
                     break;
                 case Phase.PERSISTENCE_CALLBACK:
+                    faster.lastVersion =
```
badrishc, line 224: this looks like an incorrect line.

cs/src/core/Index/Synchronization/HybridLogCheckpointTask.cs
```diff
@@ -12,13 +12,15 @@ namespace FASTER.core
     /// </summary>
     internal abstract class HybridLogCheckpointOrchestrationTask : ISynchronizationTask
     {
+        private long lastVersion;
```
badrishc, line 15: shouldnt this be initialized to -1 so that LastCheckpointedVersion will return -1 in case there are no checkpoints.

cs/test/RecoveryChecks.cs
```diff
@@ -558,6 +558,7 @@ private async ValueTask IncrSnapshotRecoveryCheck(ICheckpointManager checkpointM
                 s1.Upsert(key, key+1);
             }
 
+            var version1 = fht1.SystemState.version;
```
badrishc, line 561: Prefer to use the new public APis in testcases.

cs/src/core/Index/Synchronization/HybridLogCheckpointTask.cs
tli2: I thought FASTER versions start at 1? So 0 is no checkpoint

PR #550, opened 2021-08-23
Fix bug in AcquireLatchX

qtcwt:
Check version of traced record rather than of a bucket's tail record while deciding to update in place or COW.

Issue: Previous InVersionNew() only checks the version of the tail record in a bucket to decide whether to update in place or COW. However, when two records of different keys share the same bucket, the one in version v+1 may trick updates on another key whose version is v to incorrectly think that this key has arrived v+1 and update in place. 
Note that this might be a very rare case--two updates on different keys that collide are happening at the same time and the state machine is shifting. But it still can occur.


qtcwt:
In this PR I played safely for PREPARE phase: InVersionNew still checks the tail record, and will go pending if any key of the bucket has arrived version v+1. This might be unnecessary (see [here](https://github.com/microsoft/FASTER/pull/550/files#diff-8ccf79ac443c973bdf01265d9906728ed9de3803f45dbbc02071efe6ec790fe6R476)) and please let me know if updating records in version v while the bucket arrives v+1 is actually benign.

badrishc:
What about the case where:
1. Session is in v (prepare phase)
2. Bucket is also in v
3. Record being updated is in v+1

Looks like we will wrongly update the record in place

badrishc:
> In this PR I play safe for PREPARE phase: InVersionNew still checks the tail record, and will go pending if any key of the bucket has arrived version v+1. This might be unnecessary (see here) and please let me know if updating records in version v while their bucket arrives v+1 is actually benign.

What you have right now (no change) is correct for PREPARE phase. It is better to force the thread to shift rather than update v in place (even though it is not incorrect to do so).

badrishc:
One comment (via @tli2): suggest instead of having overloaded InVersionNew() calls, we actually separate out a CheckEntryVersionNew() and CheckBucketVersionNew() so it is very clear from context.

qtcwt:
> What about the case where:
> 
> 1. Session is in v (prepare phase)
> 2. Bucket is also in v
> 3. Record being updated is in v+1
> 
> Looks like we will wrongly update the record in place.
> 
> Edit: I think this won't happen since chains are monotonically versioned.

Right, this won't happen. If the record being updated is in v+1, then the tail of the bucket must be v+1. This case, the session in v (prepare phase) will create pending context and return CPR_SHIFT_DETECTED.

> What you have right now (no change) is correct for PREPARE phase. It is better to force the thread to shift rather than update v in place (even though it is not incorrect to do so).

Cool. I'll keep it that way then.

> One comment (via @tli2): suggest instead of having overloaded InVersionNew() calls, we actually separate out a CheckEntryVersionNew() and CheckBucketVersionNew() so it is very clear from context.

Good idea. Will do in the next commit.

PR #554, opened 2021-09-01
SpanByteClient (FASTER Remote)

acqs:
FASTER Remote client implementation of the SpanByte, which yields ~75-90% performance improvement in VariableLenClient sample (see variable length sample project).  The following are the changes related to each of the respective projects:

1. **Faster.client** - SpanByte, SpanByteMemory and IntExclusiveLocker types duplicated in Faster.client (from Faster.core) following the practice of duplication with Faster.core.Status to avoid the dependence on Faster.core.  FasterKVClient class updated to include two new "NewSession" methods.
2. **Faster.client/Faster.server** - SpanByteClientSerializer duplicated in Faster.client (from Faster.server) as it's unclear why they type is in the server project.  Assumption is that it may be relevant for inter-node communication for DPR so it was not moved.
3. VariableLengthClient - SpanByteBenchmark class added replicating the MemoryBenchmark class for SpanByteClient
3. **Faster.remote.test** - SpanByteClient, SpanByteClientTests and SpanBytePubSub test classes added following the structure of the VariableLenClient tests.


badrishc:
Nice, didn't expect client side to be impacted this much by using SpanByte (vs. Memory).

Not a topic of this PR, but we need to eventually refactor such common code into a shared project used by both client/server and core. @TedHartMS

acqs:
The benchmark outputs in VarLenClient are labeled.  SpanByte runs in roughly ~55-60% of the time it takes with Memory...

Total time: 6708ms for 10000000 gets with Memory
Total time: 3859ms for 10000000 gets with SpanByte

acqs:
Badrish,

Please take a look at the VariableLengthClient\SpanByteBenchmarks.cs file if you can spare a moment.  The ReadCompletionCallBack successfully returns 139,999 times (10,000 records per batch.  9,999 records successfully read 14 times. record 10,000 fails on iteration 14...ie succeeds 13 times prior).

Note, that I have a separate local branch with the client referencing Faster.core directly for SpanByte, SBMemory, etc.  I don't experience the problem with the referenced compilation, but I'm 99% certain it's unrelated.  I ran a diff and everything else is identical, which is the only reason I mention it.

This issue isn't caught by SpanByteTests because it only reads 100 times.

image

acqs:
Bytes from MemoryPool in SpanByteClientSerializer were not zero-initialized after on the 141K (ie key/value=="0" for first record of 10K batch with range 0-9999).  On iteration 141K, SpanByteClientSerializer.ReadOut() was returning "0999" (ie 3 additional ASCII bytes filled "57").  For some reason, ReadCompletionCallBack wasn't being triggered which led to the read performance of SpanByteBenchmark appearing dramatically better than reality.  I modified the SpanByteClientSerializer in Faster.client to resolve the issue.  Actual read benchmark metrics below from the last commit.  The performance gap appears to increase with larger payload sizes.

Note, I did not update the Faster.server copy of the SpanByteClientSerializer.

**RELEASE: ~10% faster**
Total time: 2442ms for 10000000 gets with Memory
Total time: 2218ms for 10000000 gets with SpanByte

**DEBUG: ~5% faster**
Total time: 7155ms for 10000000 gets with Memory                                                                                                                                                                                                                                                           Total time: 6827ms for 10000000 gets with SpanByte

badrishc:
I don't think the issue is with SpanByteClientSerializer, rather there is a more fundamental problem with using SpanByte on the client side. The way client works is that it stores the pending operations in a queue for later retrieval (when responses come in from the server side): `readrmwQueue.Enqueue((key, input, output, userContext)`

However, SpanByte is not safe to store in this queue because it points to memory locations that are supposed to remain fixed.

In fact, in your code, you have:

```cs
                Span<byte> bytes = Encoding.ASCII.GetBytes(k.ToString());
                fixed (byte* b = bytes)
                {
                    getBatch[k] = SpanByte.FromPointer(b, bytes.Length);
                }
```

So getBatch contents are no longer safe to use even in the session.Read call, and they are not safe to keep in the pending queue.

To reiterate, SpanByte needs to always point to fixed or pinned memory, making it a poor fit in this scenario.

acqs:
I hadn't noticed the local queue batching mechanism prior to your message.  Makes sense now.

Sincerely appreciate the explanation.  Apologies for the wasted time.

badrishc:
No worries, thanks for experimenting with this!

acqs:
Great work with FASTER, Badrish!   I haven't had time to kick the tires on the codebase thoroughly, but it seems safe to say, "hats off!"  If you smooth the corners on the (obvious) legacy tech debt, and put the necessary constructs in place for plug-and-play "mainstream developer" consumption (i.e. dumbed-down API, default production-ready persistence, and first-class support OOB support for non-bittables...the sky's the limit!

With respect to non-bittables,  I wrote a serializer (inspired to your project) that's up to ~20x faster than MessagePack for large complex types (including nullables) and ~17x faster than Protobuf.

image

I'm leveraging it to index type parameter for high performance querying.  I'd be happy to contribute it to the project.  If you're interested.  Shoot me an email to setup a call to discuss...adamtocohen@gmail.com.

P.S.  Great work on Trill as well.  I'm sure you've considered it, but FASTER+Trill could be a game changer.  Keep it up!

badrishc:
Hi, this sounds fantastic. We have a sample in playground that uses MessagePack [here](https://github.com/microsoft/FASTER/blob/master/cs/playground/AsyncStress/SerializedFasterWrapper.cs). That actually offers a dumbed-down API that is closer to what you are thinking of, I think. It would be interesting to have a version of SerializedFasterWrapper that uses your serializer. Yes, your contribution in this space would be much appreciated!

PR #564, opened 2021-10-04
[C#] Allow users to perform multiple reads under same epoch

badrishc:
…epoch.


TedHartMS (commented):
Needs more comments describing the sequence, and a sample

PR #582, opened 2021-10-21
[C++] Fix unit test

wanghenshui:
bucket(chunk * kGcHashTableChunkSize + idx) cause compact_lookup_test segfault 

also, `assert(function)` is not working in release mode(-DNDEBUG)


badrishc:
The PR https://github.com/microsoft/FASTER/pull/629 by @kkanellis we think solves this issue in the right way. Can you verify that it works for you?

wanghenshui:
> The PR #629 by @kkanellis we think solves this issue in the right way. Can you verify that it works for you?

fixed

PR #588, opened 2021-10-29
[C#] Checkpoint, Recovery, RecordInfo updates

badrishc:
* Use only 1 bit for newVersion in RecordInfo, for CPR. Semantics are that sessions in v+1 will create records in fuzzy region with this bit set. Recovery only scans fuzzy region to elide such records.
* Add 1 Filler bit for future use with fillers in records
* Increase read lock bits to 6 (64 parallel readers)
* Use dirty bit for incremental checkpoints

### DIFF
--- a/cs/src/core/Index/Common/Contexts.cs
+++ b/cs/src/core/Index/Common/Contexts.cs
@@ -36,7 +36,7 @@ namespace FASTER.core
 
     internal class SerializedFasterExecutionContext
     {
-        internal int version;
+        internal long version;
         internal long serialNum;
         internal string guid;
 
@@ -56,7 +56,7 @@ namespace FASTER.core
         public void Load(StreamReader reader)
         {
             string value = reader.ReadLine();
-            version = int.Parse(value);
+            version = long.Parse(value);
 
             guid = reader.ReadLine();
             value = reader.ReadLine();
@@ -78,7 +78,7 @@ namespace FASTER.core
 
             // Some additional information about the previous attempt
             internal long id;
-            internal int version;
+            internal long version;
             internal long logicalAddress;
             internal long serialNum;
             internal HashBucketEntry entry;
@@ -248,7 +248,7 @@ namespace FASTER.core
     /// </summary>
     public struct HybridLogRecoveryInfo
     {
-        const int CheckpointVersion = 2;
+        const int CheckpointVersion = 3;
 
         /// <summary>
         /// Guid
@@ -261,11 +261,11 @@ namespace FASTER.core
         /// <summary>
         /// Version
         /// </summary>
-        public int version;
+        public long version;
         /// <summary>
         /// Next Version
         /// </summary>
-        public int nextVersion;
+        public long nextVersion;
         /// <summary>
         /// Flushed logical address; indicates the latest immutable address on the main FASTER log at recovery time.
         /// </summary>
@@ -318,7 +318,7 @@ namespace FASTER.core
         /// </summary>
         /// <param name="token"></param>
         /// <param name="_version"></param>
-        public void Initialize(Guid token, int _version)
+        public void Initialize(Guid token, long _version)
         {
             guid = token;
             useSnapshotFile = 0;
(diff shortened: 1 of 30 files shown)


cs/src/core/Index/Common/RecordInfo.cs
```diff
@@ -228,26 +238,33 @@ public bool Dirty
...
             {
-                if (value) word &= ~kValidBitMask;
-                else word |= kValidBitMask;
+                if (value) word |= kFillerBitMask;
+                else word &= ~kFillerBitMask;
             }
         }
 
-        public int Version
+        public bool NewVersion
```
tli2, line 251: nit: InNewVersion?

cs/src/core/Index/Common/RecordInfo.cs
```diff
@@ -43,33 +43,28 @@ public struct RecordInfo
...
-        const long kVersionMaskInWord = ((1L << kVersionBits) - 1) << kVersionShiftInWord;
-        internal const long kVersionMaskInInteger = (1L << kVersionBits) - 1;
+        const long kFillerBitMask = 1L << kFillerBitOffset;
+        const long kNewVersionBitMask = 1L << kNewVersionBitOffset;
 
         [FieldOffset(0)]
         private long word;
 
-        public static void WriteInfo(ref RecordInfo info, int checkpointVersion, bool tombstone, bool invalidBit, long previousAddress)
+        public static void WriteInfo(ref RecordInfo info, bool newVersion, bool tombstone, bool dirty, long previousAddress)
```
tli2, line 60: nit: InNewVersion?

cs/src/core/Index/Common/RecordInfo.cs
```diff
@@ -228,26 +238,33 @@ public bool Dirty
             }
         }
 
-        public bool Invalid
```
tli2, line 231: What is Filler?

cs/src/core/Index/Common/Contexts.cs
```diff
@@ -221,6 +221,8 @@ public async ValueTask WaitPendingAsync(CancellationToken token = default)
                     await readyResponses.WaitForEntryAsync(token).ConfigureAwait(false);
             }
 
+            public bool NewVersion => phase < Phase.REST;
```
tli2, line 224: nit: InNewVersion?

cs/src/core/Index/FASTER/FASTERImpl.cs
```diff
@@ -28,26 +28,38 @@ public unsafe partial class FasterKV<Key, Value> : FasterBase, IFasterKV<Key, Va
...
-        /// <param name="sessionCtx"></param>
         /// <returns></returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private bool CheckBucketVersionNew<Input, Output, Context>(ref HashBucketEntry entry, FasterExecutionContext<Input, Output, Context> sessionCtx)
+        private bool CheckBucketVersionNew(ref HashBucketEntry entry)
         {
             // A version shift can only in an address after the checkpoint starts, as v_new threads RCU entries to the tail.
             if (entry.Address < _hybridLogCheckpoint.info.startLogicalAddress) return false;
+
             // Otherwise, check if the version suffix of the entry matches v_new.
```
tli2, line 46: remove outdated comment?

tli2 (approved):
LGTM

cs/src/core/Index/Common/RecordInfo.cs
badrishc: We will use Filler to track unused space in the main log, which might be present if we resized a value from say 100 down to 60. This will allow us to in-place update the value back from 60 to say 80.

badrishc: Specifically, the plan is: if Filler = true, then the 8 bytes following Value (or Key in case of Tombstone) will denote the size of the filler space, i.e. the space in bytes before the start of the next RecordInfo on that page.

cs/src/core/Index/FASTER/FASTERImpl.cs
badrishc: I made one minor change to CheckBucketVersionNew in checkin, pls review.

PR #593, opened 2021-11-14
[C#] FasterLog v2 - misc changes, removed OverWrite commits

badrishc:


cs/src/core/Index/CheckpointManagement/DeviceLogCommitCheckpointManager.cs
```diff
@@ -180,36 +147,11 @@ public byte[] GetCommitMetadata(long commitNum)
                 body = writePad;
             else
                 ReadInto(device, 0, out body, size + sizeof(int));
-            
-            if (!overwriteLogCommits)
```
tli2, line 184: don't we need to dispose of the device when we are done?

cs/src/core/Index/CheckpointManagement/DeviceLogCommitCheckpointManager.cs
```diff
@@ -429,6 +372,35 @@ public void OnRecovery(Guid indexToken, Guid logToken)
             }
         }
 
+        /// <inheritdoc />
+        public void OnRecovery(long commitNum, bool purgeEarlierCommits)
+        {
+            if (removeOutdated || purgeEarlierCommits)
```
tli2, line 378: This block is slightly confusing to me with many if conditionals: maybe just flatten it into:
```
if (removeOutdated) {
  ... 
}
else if (purgeEarlierCommits) {
 ...
return;
}
```

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -1166,28 +1163,27 @@ private void UpdateTailCallback(long tailAddress)
         }
 
         
-        // TODO(Tianyu): Will we ever need to recover to a specific commit for read-only cases?
         /// <summary>
-        /// Synchronously recover instance to FasterLog's latest commit, when being used as a readonly log iterator
+        /// Synchronously recover instance to FasterLog's latest valid commit, when being used as a readonly log iterator
         /// </summary>
-        public void RecoverReadOnly()
+        public void RecoverReadOnly(bool purgeEarlierCommits = false)
```
tli2, line 1169: i feel like this might be a bit weird to allow readonly clients to purge commits? Definitely smells like a write operation to me. This is especially weird because I don't think the non-readonly FasterLog has a similar option.

cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -1232,39 +1228,44 @@ private bool LoadCommitMetadata(long commitNum, out FasterLogRecoveryInfo info)
...
+                    }
                 }
                 catch { }
             }
 
             // Only in fast commit mode will we potentially need to recover from an entry in the log
             if (fastCommitMode)
             {
-                // Shut up safe guards, I know what I am doing
+                // Disable safe guards temporarily
```
tli2, line 1254: haha

cs/src/core/Index/FasterLog/FasterLogRecoveryInfo.cs
```diff
@@ -66,7 +69,7 @@ public void Initialize(BinaryReader reader)
                 checkSum = reader.ReadInt64();
                 BeginAddress = reader.ReadInt64();
                 UntilAddress = reader.ReadInt64();
-                if (version == 1)
```
tli2, line 69: nit: why change to > 0? I feel like with version numbers we want to be as future-proof as possible.

cs/src/core/Index/FasterLog/FasterLogRecoveryInfo.cs
```diff
@@ -75,39 +78,56 @@ public void Initialize(BinaryReader reader)
             {
                 throw new FasterException("Unable to recover from previous commit. Inner exception: " + e.ToString());
             }
-            if (version != 0 && version != 1)
+            if (version < 0 || version > FasterLogRecoveryVersion)
```
tli2, line 81: same as above

cs/src/core/Index/FasterLog/FasterLog.cs
badrishc: The real question is, how then can we purge commits. Having the writer purge commits makes it impossible for readers to consistently read the latest as it gets deleted after reader discovers it and before it reads.  Or worse,  exceptions thrown if writer deletes when reader reading.

cs/src/core/Index/CheckpointManagement/DeviceLogCommitCheckpointManager.cs
badrishc: "using var device" should handle it

cs/src/core/Index/FasterLog/FasterLog.cs
tli2: The same problem is present with this version if we have multiple readers right?

The dumb and easy fix in this case seems to be have readers retry on exceptions thrown -- purged commits will always be replaced by a later commit.

Also, I am not certain about this: do read only consumers of FasterLog automatically catch up to newly committed tails, or do they need to continuously invoke RecoverReadOnly()? If latter, this might be a bad idea for fastcommits due to double scanning.

badrishc: > The same problem is present with this version if we have multiple readers right?

correct, that is why it is exposed to the API so that only the last reader would invoke it with purge=true.

> do read only consumers of FasterLog automatically catch up to newly committed tails, or do they need to continuously invoke RecoverReadOnly()?

It is the latter. Each call to RecoverReadOnly moves the reader's view forward as it accesses the latest commit.dat.

>  this might be a bad idea for fastcommits

Yes I was going to bring this up w.r.t. FastCommit. Our RecoverReadOnly in this case just needs to keep reading forward until the next commit record, backing up if corrupt record or no commit record found, instead of relying on commit.dat.

badrishc: > The dumb and easy fix in this case seems to be have readers retry on exceptions thrown -- purged commits will always be replaced by a later commit.

This might actually never happen (loop forever), because commits and deletions are happening so fast that the reader retries may never successfully capture a commit file. Also, note that the exception happens in the writer (unable to delete -- when we try to delete a file that the reader has open).

cs/src/core/Index/FasterLog/FasterLogRecoveryInfo.cs
badrishc: the goal was that if we updated the version to say 2, the rest of the code would work correctly out of the box as baseline. then we would just make the changes necessary for whatever caused us to move to version 2.

cs/src/core/Index/FasterLog/FasterLog.cs
tli2: We can probably punt on this problem for this PR. I personally think this should be part of the larger producer/consumer overhaul (e.g., with explicit Complete() calls for read iterators).

In this model, readonly fasterlog is just an iterator (or a bunch) on a difference process. Whatever fix we do that should apply here too with some generalization.

tli2 (approved):

cs/src/core/Index/CheckpointManagement/DeviceLogCommitCheckpointManager.cs
badrishc: fixed

cs/src/core/Index/FasterLog/FasterLog.cs
badrishc: Sounds good, we would likely want this overhaul to be in v2 as it might be a breaking change.

PR #594, opened 2021-11-15
[C#] Modify SubscribeKVProvider to support custom functions

rohankadekodi-msr:
* Added a GetFunctions() method in SpanByteFasterKVProvider, to support custom functions with the SpanByte provider.
* Fixed a bug in SubscribeKVBroker, where it was calling Publish for normal subscriptions instead of prefix subscriptions.


badrishc:
closing to recreate without interface overhead.

PR #596, opened 2021-11-15
[C#] v2 pending flush list

badrishc:


cs/src/core/Allocator/PendingFlushList.cs
```diff
@@ -1,54 +1,43 @@
...
-            list = new PageAsyncFlushResult<Empty>[maxSize];
+            list = new();
         }
 
-        public bool Add(PageAsyncFlushResult<Empty> t)
+        public void Add(PageAsyncFlushResult<Empty> t)
         {
-            int retries = 0;
-            do
+            lock (list)
```
tli2, line 19: Perf idea: let's reuse the original performant code path with a fixed size array --- if array full, grow the array under critical section. For concurrency, can use something similar to SimpleVersionScheme in libDPR to ensure minimal overhead on the normal code path.

cs/src/core/Allocator/PendingFlushList.cs
```diff
@@ -1,54 +1,43 @@
...
-                    {
-                        if (Interlocked.CompareExchange(ref list[i], t, default) == default)
-                        {
-                            return true;
-                        }
-                    }
-                }
-            } while (retries++ < maxRetries);
-            return false;
+                list.AddFirst(t);
```
tli2, line 21: just checking: we basically don't have any ordering guarantees within the pending flush list right? 

Is there maybe some value in ensuring order? Normally, we don't flush a random range to disk, so only truly concurrent flushes will result in out-of-order entries in the pending flush list. In that case, whatever saving we have on the add path will arguably be negated on the remove path. What about just biting the bullet and ensure ordering from the start?

cs/src/core/Allocator/PendingFlushList.cs
badrishc: Growing is fine, but then we would have to shrink at some point. Also, the growability implies we still have to take the lock every time.

cs/src/core/Allocator/PendingFlushList.cs
badrishc: Ensuring order would be useful. The entries are of the form [0, 23], [24, 39], [40, 100] and so on. But keeping these ordered will make this a O(log N) structure (interval tree).

cs/src/core/Allocator/PendingFlushList.cs
badrishc: I was also thinking of a LinkedList, where each element is the performant array from earlier. Kind of like the ElasticCircularBuffer ...

cs/src/core/Allocator/PendingFlushList.cs
tli2: I was thinking about a much simpler approach: if you have [0, 23], and you see [40, 100] --- [24,39] must have been issued from a different thread, we can just spin wait until that's in before enqueuing [40, 100]

cs/src/core/Allocator/PendingFlushList.cs
tli2: We don't really need to shrink here? Even a thousand pending flush entry is nothing in terms of mem size. 

For the lock --- we can use epoch + one atomic variable to avoid a shared lock on the common path. SimpleVersionScheme does this.

Issue #601, opened 2021-11-23
[C#] FasterLogIterator.CompleteUntilRecordAt() should not verify checksum of the record, even if checksumming is enabled


mito-csod:
CompleteUntilRecordAt() currently fails occasionally due to checksum failures. This is because the AsyncGetHeaderOnlyFromDiskCallback method reads only enough data in to capture the header. The solution is to never verify checksum in GetRecordLengthAndFree()--this should be "safe" since this method is not used to retrieve records themselves.

badrishc:
Makes sense, merged PR.

PR #602, opened 2021-11-23
Disable checksum verification in FasterLogIterator.CompleteUntilRecordAt()

mito-csod:
Addresses occasional checksum failures as described in #601.


badrishc:
Merged, thanks.

PR #606, opened 2021-12-01
[C#] Add static option to unpin objects when returning to SectorAlignedMemoryPool

badrishc:
To enable, do this at program entry point:

`SectorAlignedBufferPool.UnpinOnReturn = true;`


badrishc:
Closing for now, we will revisit this option in future.

badrishc:
Reopening as we found this setting allows better GC behavior and possibly improved CPU/throughput in some scenarios. It will still be turned off as default for now.

PR #608, opened 2021-12-08
[C#] Add TentativeHeadAddress to allocator (v2)

badrishc:
Added `TentativeHeadAddress` - we first shift this, and when all threads agree (via epoch bump), we can do some pre-closing work such as eviction. Then we can shift the actual `HeadAddress`, finally closing pages when all threads agree on this (via epoch bump).

### DIFF
--- a/cs/src/core/Allocator/AllocatorBase.cs
+++ b/cs/src/core/Allocator/AllocatorBase.cs
@@ -147,6 +147,12 @@ namespace FASTER.core
         public long SafeHeadAddress;
 
         /// <summary>
+        /// Tentative head address. Threads do not take any record locks earlier than this point.
+        /// Records earlier than this address undergo eviction, before HeadAddress is moved.
+        /// </summary>
+        public long TentativeHeadAddress;
+
+        /// <summary>
         /// Flushed until address
         /// </summary>
         public long FlushedUntilAddress;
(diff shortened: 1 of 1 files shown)


cs/src/core/Allocator/AllocatorBase.cs
```diff
@@ -1334,29 +1356,7 @@ private void PageAlignedShiftReadOnlyAddress(long currentTailAddress)
...
-            long pageAlignedTailAddress = currentTailAddress & ~PageSizeMask;
-            long desiredHeadAddress = (pageAlignedTailAddress - HeadOffsetLagAddress);
-
-            long newHeadAddress = desiredHeadAddress;
-            if (currentFlushedUntilAddress < newHeadAddress)
-            {
-                newHeadAddress = currentFlushedUntilAddress;
-            }
-            newHeadAddress &= ~PageSizeMask;
-
```
TedHartMS, line 1350: newHeadAddress &= ~PageSizeMask;
is not in the new call-through version. It seems needed if wew set newHeadAddress = currentFlushedUntilAddress;

badrishc: I saw that earlier, but it is a benign change. The reason we have PageAligned is to avoid many small head movements as the tail grows, and this version satisfies that requirement.

TedHartMS (approved):

PR #611, opened 2021-12-09
[C#] minor cleanup of fasterlog v2

badrishc:


cs/src/core/Index/FasterLog/FasterLog.cs
```diff
@@ -189,7 +189,7 @@ private FasterLog(FasterLogSettings logSettings, bool oldCommitManager)
             fastCommitMode = logSettings.FastCommitMode;
 
             ongoingCommitRequests = new Queue<(long, FasterLogRecoveryInfo)>();
-            commitPolicy = logSettings.CommitPolicy ?? new DefaultCommitPolicy();
+            commitPolicy = logSettings.LogCommitPolicy ?? LogCommitPolicy.Default();
```
tli2, line 192: No need to switch here because already initialized default in FasterLogSettings

tli2 (approved):
LGTM

cs/src/core/Index/FasterLog/FasterLog.cs
badrishc: just an additional level of protection in case user sets it to null for some reason.

PR #623, opened 2021-12-19
[C#] FasterLog v2 iteration fix

badrishc:


cs/src/core/FasterLog/FasterLog.cs
```diff
@@ -120,12 +120,21 @@ public sealed class FasterLog : IDisposable
...
         public FasterLog(FasterLogSettings logSettings)
+            : this(logSettings, logSettings.TryRecoverLatest)
+        { }
+
+        /// <summary>
+        /// Create new log instance
+        /// </summary>
+        /// <param name="logSettings">Log settings</param>
+        /// <param name="syncRecover">Recover synchronously</param>
+        private FasterLog(FasterLogSettings logSettings, bool syncRecover)
```
tli2, line 137: why not have syncRecover = true to cover the earlier case as well?

cs/src/core/FasterLog/FasterLog.cs
```diff
@@ -167,7 +176,8 @@ public FasterLog(FasterLogSettings logSettings)
             commitPolicy.OnAttached(this);
 
             tolerateDeviceFailure = logSettings.TolerateDeviceFailure;
-            if (logSettings.TryRecoverLatest)
+
+            if (syncRecover)
```
tli2, line 180: Need to check recover latest here as well?

If not, maybe we don't need to separate flags for recover latest and recoversync?

tli2 (approved):

cs/src/core/FasterLog/FasterLog.cs
badrishc: The private constructor needed to differentiate itself from the public one, so I made it contingent on recoverSync. It does not look at log settings, instead depends on caller.

cs/src/core/FasterLog/FasterLog.cs
badrishc: We want public constructor to be clean with only one argument, the settings.

cs/src/core/FasterLog/FasterLog.cs
badrishc: The CreateAsync call invokes this private constructor with this argument false, and takes care of async recovery itself.

cs/src/core/FasterLog/FasterLog.cs
tli2: Ok

PR #630, opened 2022-01-20
[C#] Add WriteReason parameter to SingleWriter

TedHartMS:


cs/benchmark/Functions.cs
```diff
@@ -51,13 +51,7 @@ public bool ConcurrentReader(ref Key key, ref Input input, ref Value value, ref
...
         // Upsert functions
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void SingleWriter(ref Key key, ref Input input, ref Value src, ref Value dst, ref Output output, ref RecordInfo recordInfo, long address)
-        {
-            dst = src;
-        }
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void CopyWriter(ref Key key, ref Value src, ref Value dst, ref RecordInfo recordInfo, long address)
+        public void SingleWriter(WriteReason reason, ref Key key, ref Input input, ref Value src, ref Value dst, ref Output output, ref RecordInfo recordInfo, long address)
```
badrishc, line 54: I think we want the API to be:

```cs
SingleWriter(ref Key key, ref Input input, ref Value src, ref Value dst, ref Output output, ref RecordInfo recordInfo, long address, WriteReason reason)
```

cs/src/core/Index/FASTER/FASTER.cs
```diff
@@ -54,6 +54,24 @@ public enum ReadFlags
         SkipCopyReads = SkipReadCache | SkipCopyToTail,
     }
 
+    /// <summary>
+    /// The reason a SingleWriter was performed
+    /// </summary>
+    public enum WriteReason
```
badrishc, line 60: move to its own file for cleanliness?

TedHartMS: done

cs/benchmark/Functions.cs
TedHartMS: done

PR #639, opened 2022-01-29
[C#] IEntry interface for FasterLog

tli2:
By specifying an IEntry interface, users can now serialize objects directly onto FasterLog when enqueuing for performance.


cs/src/core/FasterLog/FasterLog.cs
```diff
@@ -19,6 +19,22 @@ namespace FASTER.core
...
     public sealed class FasterLog : IDisposable
     {
+        /// <summary>
+        ///  Represents a entry that can be serialized directly onto FasterLog when enqueuing
+        /// </summary>
+        public interface IEntry
+        {
+            /// <summary></summary>
+            /// <returns> the size in bytes after serialization onto FasterLog</returns>
+            public int SerializedLength();
```
iamcarbon, line 29: should this be GetSerializedLength() or SerializedLength { get; } ?

badrishc: SerializedLength { get; } looks like a good option.

PR #641, opened 2022-02-01
[C#] Support int session ID for FasterKV

badrishc:
* Session ID (int) and optional session name (string) are supported
* Ensure that checkpoints write out recoverable sessions (i.e., sessions that were restored from the previous checkpoint, but not yet explicitly resumed) to the commit info, even if these sessions were not explicitly resumed after recovery.
* Add APIs to iterate as well as dispose recoverable sessions in store, after calling `Recover`
* Correctly handle active sessions that are disposed during a checkpoint (by waiting until the checkpoint completes before removing them)

### DIFF
--- a/cs/benchmark/Functions.cs
+++ b/cs/benchmark/Functions.cs
@@ -25,9 +25,9 @@ namespace FASTER.benchmark
         {
         }
 
-        public void CheckpointCompletionCallback(string sessionId, CommitPoint commitPoint)
+        public void CheckpointCompletionCallback(int sessionID, string sessionName, CommitPoint commitPoint)
         {
-            Debug.WriteLine("Session {0} reports persistence until {1}", sessionId, commitPoint.UntilSerialNo);
+            Debug.WriteLine("Session {0} reports persistence until {1}", sessionID, commitPoint.UntilSerialNo);
         }
 
         // Read functions
--- a/cs/benchmark/Functions.cs
+++ b/cs/benchmark/Functions.cs
@@ -27,7 +27,7 @@ namespace FASTER.benchmark
 
         public void CheckpointCompletionCallback(int sessionID, string sessionName, CommitPoint commitPoint)
         {
-            Debug.WriteLine("Session {0} reports persistence until {1}", sessionID, commitPoint.UntilSerialNo);
+            Debug.WriteLine($"Session {sessionID} ({(sessionName ?? "null")}) reports persistence until {commitPoint.UntilSerialNo}");
         }
 
         // Read functions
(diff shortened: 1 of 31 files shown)


cs/benchmark/Functions.cs
```diff
@@ -25,9 +25,9 @@ public void DeleteCompletionCallback(ref Key key, Empty ctx)
         {
         }
 
-        public void CheckpointCompletionCallback(string sessionId, CommitPoint commitPoint)
+        public void CheckpointCompletionCallback(int sessionID, string sessionName, CommitPoint commitPoint)
         {
-            Debug.WriteLine("Session {0} reports persistence until {1}", sessionId, commitPoint.UntilSerialNo);
+            Debug.WriteLine("Session {0} reports persistence until {1}", sessionID, commitPoint.UntilSerialNo);
```
tli2, line 30: Nit: Include name in debug print as well?

cs/src/core/Index/Common/Contexts.cs
```diff
@@ -38,15 +38,17 @@ internal class SerializedFasterExecutionContext
...
+        internal string sessionName;
 
         /// <summary>
         /// </summary>
         /// <param name="writer"></param>
         public void Write(StreamWriter writer)
         {
             writer.WriteLine(version);
-            writer.WriteLine(guid);
+            writer.WriteLine(sessionName);
```
tli2, line 50: Does this work correctly with null sessionNames?

tli2 (commented):
Just took a quick look. Will look in more detail later.

cs/src/core/Index/Common/Contexts.cs
badrishc: fixed

cs/benchmark/Functions.cs
badrishc: done

cs/remote/samples/FixedLenServer/Types.cs
```diff
@@ -64,8 +64,8 @@ public struct Functions : IFunctions<Key, Value, Input, Output, long>
 
         public void DeleteCompletionCallback(ref Key key, long ctx) { }
 
-        public void CheckpointCompletionCallback(string sessionId, CommitPoint commitPoint)
-            => Debug.WriteLine("Session {0} reports persistence until {1}", sessionId, commitPoint.UntilSerialNo);
+        public void CheckpointCompletionCallback(int sessionID, string sessionName, CommitPoint commitPoint)
+            => Debug.WriteLine("Session {0} reports persistence until {1}", sessionID, commitPoint.UntilSerialNo);
```
TedHartMS, line 69: SessionName here?

Issue #640, opened 2022-01-30
FasterLog based Pubsub hangs if log file deleted


vangarp:
In the code below, I expect that after I delete the log file and restart the producer and consumer that my consumer would obviously miss everything stored in the log file in iteration 1, but it should be able to consume the messages produced during the 2nd iteration. However, it seems like the consumer completely hangs and has no way to get out. Even if I wire up cancellation tokens, the iterations just keep on cancelling out at the iteration.WaitAsync() call. How can we get out of this state?

Version: 1.9.10

```
namespace FasterLogPlayground
{
    using System;
    using System.Collections.Generic;
    using System.IO;
    using System.Runtime.CompilerServices;
    using System.Text;
    using System.Threading.Tasks;
    using FASTER.core;

    internal class Program
    {
        private static IDevice device;
        private static FasterLog log;
        private static FasterLogScanIterator iterator;

        static async Task Main(string[] args)
        {
            // Produce some logs and consume them partially
            device = Devices.CreateLogDevice("hlog.log", useIoCompletionPort: true);
            log = new FasterLog(new FasterLogSettings { LogDevice = device, MemorySizeBits = 26, PageSizeBits = 20, MutableFraction = 0.5, SegmentSizeBits = 20 });
            iterator = log.Scan(log.BeginAddress, long.MaxValue, "logtest1", true, ScanBufferingMode.DoublePageBuffering, true);

            TaskCompletionSource<bool> firstIterationTaskCompletionSource = new TaskCompletionSource<bool>();
            Task.Run(() => CommitterAsync(log, TimeSpan.FromMilliseconds(100), firstIterationTaskCompletionSource));
            var tasks = new List<Task>();
            tasks.Add(ProducerAsync(log, "Message ", 100));
            tasks.Add(ConsumerAsync(iterator, log, "Consumer1", 10));
            await Task.WhenAll(tasks).ConfigureAwait(false);
            firstIterationTaskCompletionSource.SetResult(true);

            // Give some delay for commit to complete
            await Task.Delay(500).ConfigureAwait(false);

            iterator.Dispose();
            log.Dispose();
            device.Dispose();

            // Delete the log file
            File.Delete("hlog.log.0");

            // Now try to produce and consume again.
            // Expected behavior: My expectation is that we will recover and start to consume whatever is produced starting at this point.
            // Actual behavior: The producer produces everything, but consumer is stuck forever
            device = Devices.CreateLogDevice("hlog.log", useIoCompletionPort: true);
[...]

PR #642, opened 2022-02-01
[C#] Throw exception on page read/flush error during restore

badrishc:
Fix https://github.com/microsoft/FASTER/issues/640

### DIFF
--- a/cs/src/core/Index/Recovery/Recovery.cs
+++ b/cs/src/core/Index/Recovery/Recovery.cs
@@ -13,7 +13,7 @@ using System.Threading.Tasks;
 
 namespace FASTER.core
 {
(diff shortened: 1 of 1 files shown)


badrishc:
This PR makes FasterLog recovery logic throw an exception if there is a page read exception from IDevice. This exception is caught by upstream FasterLog recovery logic, which will in turn not recover to the specified commit. Thus, we will start with a clean unrecovered slate in this case, which is the best we can do.

In v2, this exception can be caught by user, by explicitly calling fasterlog.Recover() instead of setting logSettings.TryRecoverLatest. Then, the user can do any other custom repair.

tli2 (approved):
LGTM

PR #650, opened 2022-02-12
[C#] [WIP] Add Revivification

TedHartMS:
Allow tombstoned records to be reactivated. As part of this, we carry the used vs. full length of the value space of a record, allowing concurrent updates to shrink and re-grow the value within the same record.


docs/_docs/31-fasterkv-revivification.md
```diff
@@ -0,0 +1,68 @@
...
+    - The smallest bin is of size 16, allowing reuse of records large enough to hold an integer key and value. 
+
+#### Fixed vs. VarLen
+For non-variable-length types, the record size is fixed, so the FreeRecordPool has only the one bin for it. Otherwise, it has the full range of variable-length bins.
+
+#### Enqueueing
+Bins hold records up to their maxium size, which is a power of 2. Enqueuing will enqueue a record into the "smallest" bin it fits into.
+
+#### Dequeueing
+Records stored in a bin have lengths that are *between* the lowest and highest sizes of a bin, *not* the highest size of the bin. Therefore, for performance we have to retrieve from the next-highest bin; e.g. 48 will come from the [64-127] bin because the [32-63] bin might not have anything larger than 42, so we can't satisfy the request and it would take too much time to find that out before moving to the larger bin.
```
badrishc, line 46: What does dequeue do if the thread's associated partition is empty? does it check other partitions.

docs/_docs/31-fasterkv-revivification.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Revivifying a record
+The FreeRecordList is used for:
+- An Upsert, RMW, or Delete that adds a new record
+- A Pending Read or RMW that does CopyToTail (Revivification is not used for the ReadCache).
+
+If there is a record available, it is reused; otherwise, the usual BlockAllocate is done.
```
badrishc, line 53: When doing the CAS, we want to ensure that the record will point to a lower logical address. I.e., hash chains always go from higher addresses to lower addresses.

docs/_docs/31-fasterkv-revivification.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Record Length Management
+There are two lengths we are concerned with: The FullValueLength (allocated space), and the UsedValueLength within that that the application has used. These are properties of the new `UpdateInfo` structure passed by reference to `IFunctions` update callbacks for Upsert, RMW, and Delete.
+
+The used space is for a new record is equal to the full allocation space; by default, "all space is used". Note that SpanByte actually initializes itself to include space for the entire record. However, other data types may not do this.
```
badrishc, line 58: I do not get the comment about SpanByte here. Maybe clarify.

docs/_docs/31-fasterkv-revivification.md
```diff
@@ -0,0 +1,68 @@
...
+
+## FreeList
+If `FasterKVSettings.MaxFreeRecordsInBin` is nonzero, this activates the freelist.
+
+When the FreeList is active and a record is deleted, if it is at the tail of the hash chain, is not locked, and its PreviousAddress points below BeginAddress, then it can be CASed (Compare-And-Swapped) out of the hash chain. If this succeeds, it is added to the freelist.
+
+### Design of FreeList
+The FreeList hierarchy consists of:
+- The FreeRecordPool, which maintains the bins, deciding which bin should be used for Enqueue and Dequeue.
+- Multiple FreeRecordBins, including one for oversize (where the length is too long to fit into the FreeRecordHeader)
```
badrishc, line 32: Can you mention the largest bin size here?

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+ReadCompletionCallback receives the `RecordMetadata` of the record that was read.
+
+##### Return Values of IFunctions Data Operations
+IFunctions calls receive data through reference parameters to Keys and their associated Values. As well, they receive a parameter of either `ReadInfo` or `UpdateInfo`, describing additional information about the operation and its context:
+- **ReadInfo**: Contains the session Version, session type, and logical address
+- **UpdateInfo**: Contains the session Version, session type, and logical address, as well as two aspects of the value length that are important for reuse of variable-length records, either by shrinking and re-growing the value, or by revivifying a tombstoned record:
+  - **FullValueLength**: The full size of the allocated value space.
+  - **UsedValueLength**: The amount of the allocated value space that is in use. This is initially set by FASTER to the FullValueLength, and is reset by the application's IFunctions callback as appopriate.
+  - **DeleteRecord**: Always False coming in from FASTER; may be set True by the app if, for example, expiration has been detected.
```
badrishc, line 118: who sets the tombstone when this is set to True?

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+  - ***Expiration***: The app sets readInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by forcing a CopyToTail of the current record with the Tombstone set, and Read will return (!status.Found && status.CreatedRecord).
+  - ***Cancellation***: The app sets readInfo.CancelOperation True and returns false. Read() returns (!status.Found).
+- `ConcurrentReader`
+  - ***Required Length***: Irrelevant because this is not an update
+  - ***Expiration***: The app sets readInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by in-place update, and Read will return (!status.Found).
+  - ***Cancellation***: The app sets readInfo.CancelOperation True and returns false. Read() returns (!status.Found).
+
+**RMW**
+  - `InPlaceUpdater`
+    - ***Required Length***: The app sets updateInfo.FullValueLength and updateInfo.UsedValueLength to the existing lengths and returns false. FASTER will allocate a record with the required length and call `CopyUpdater`.
```
badrishc, line 138: why does the app set FullValueLength? isnt this an input to the Function and not an output.

badrishc: how is the required length provided to FASTER by the app in this case? Using GetLength or something else? Does GetLength get deprecated in this case?

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+    - ***Expiration***: The app sets updateInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by in-place update, and RMW will return (!status.Found).
```
badrishc, line 139: in case item is set to deleted, why would the app return notfound? wouldnt we have to retry and go through the initial-updater code path? the operation is not canceled in this case.

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+    - ***Cancellation***: The app sets updateInfo.CancelOperation True and returns false. RMW() returns (!status.Found).
+  - *InitialUpdater*
+    - `NeedInitialUpdate`
+      - ***Required Length***: FASTER sets updateInfo.FullValueLength and updateInfo.UsedValueLength on input to the VarLengthStructSettings result or revivification info. The app sets it to the required length if necessary. FASTER will allocate a record with the required length and call `InitialUpdater`.
+      - ***Expiration***: The app sets updateInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by continuing with InitialUpdater, and RMW will return (!status.Found && status.CreatedRecord).
```
badrishc, line 144: If NIU returns false, why do you say the RMW returns status.CreatedRecord? What record is created?

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+Additionally, the callback return values control some aspects of the operation, particularly related to:
+- The ***Required Length*** to insert or update a value exceeds the available value length
+- Record ***Expiration*** is detected
+- The application wants to ***Cancel*** the current operation
+
+**Read**:
+- `SingleReader`:
+  - ***Required Length***: Irrelevant because this is not an update
+  - ***Expiration***: The app sets readInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by forcing a CopyToTail of the current record with the Tombstone set, and Read will return (!status.Found && status.CreatedRecord).
+  - ***Cancellation***: The app sets readInfo.CancelOperation True and returns false. Read() returns (!status.Found).
```
badrishc, line 130: Read should return status.IsCanceled

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+
+Additionally, the callback return values control some aspects of the operation, particularly related to:
+- The ***Required Length*** to insert or update a value exceeds the available value length
+- Record ***Expiration*** is detected
+- The application wants to ***Cancel*** the current operation
+
+**Read**:
+- `SingleReader`:
+  - ***Required Length***: Irrelevant because this is not an update
+  - ***Expiration***: The app sets readInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by forcing a CopyToTail of the current record with the Tombstone set, and Read will return (!status.Found && status.CreatedRecord).
```
badrishc, line 129: We only need a tombstone key, i.e., no value, here at tail.

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+      - ***Expiration***: The app sets updateInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by continuing with InitialUpdater, and RMW will return (!status.Found && status.CreatedRecord).
+      - ***Cancellation***: The app sets updateInfo.CancelOperation True and returns false. RMW() returns (!status.Found).
+    - `InitialUpdater`
+      - ***Required Length***: FASTER sets updateInfo.FullValueLength and updateInfo.UsedValueLength on input to the values previously obtained.
+        - If the length is sufficient, the app sets updateInfo.UsedValueLength to the used length and returns true.
+        - Otherwise, the app sets updateInfo.UsedValueLength to the required length and returns false. RMW will return RETRY_NOW and FASTER will retry the operation.
+      - ***Expiration***: The app sets updateInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by continuing with the CAS, and RMW will return (!status.Found && status.CreatedRecord).
```
badrishc, line 150: What does it mean to expire during NIU or IU? The record does not exist to begin with, which is why NIU/IU was called, so no point adding a tombstone etc.

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,95 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+        - If the length is sufficient, the app sets updateInfo.UsedValueLength to the used length and returns true.
+        - Otherwise, the app sets updateInfo.UsedValueLength to the required length and returns false. FASTER will allocate a record with the required length and call `InitialUpdater`.
+      - ***Expiration***: The app sets updateInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by in-place update, and RMW will return (!status.Found).
+      - ***Cancellation***: The app sets updateInfo.CancelOperation True and returns false. RMW() returns (!status.Found).
+  - *CopyUpdater*
+    - `NeedCopyUpdate`
+      - ***Required Length***: FASTER sets updateInfo.FullValueLength and updateInfo.UsedValueLength on input to the VarLengthStructSettings result or revivificatgion info. The app sets it to the required length if necessary. FASTER will allocate a record with the required length and call `CopyUpdater`.
+      - ***Expiration***: The app sets updateInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by calling InitialUpdater, and RMW will return (!status.Found && status.CreatedRecord).
```
badrishc, line 161: can we allow the user to expire the data AND cancel the operation. so we are deleting the record by adding a tombstone, and we will not call IU.

docs/_docs/31-fasterkv-revivification.md
TedHartMS: It proceeds to the next bin and wraps. This is discussed earlier in the doc, but comments were added to the Enqueue and Dequeue sections for clarity.

docs/_docs/31-fasterkv-revivification.md
TedHartMS: Added to Dequeue section:

Dequeueing takes a "minimum address" parameter to maintain the invariant that hash chains point downward; when requesting a record for dequeueing, the `HashTableEntry.Address` is passed. This is either a valid lower address or an invalid address (below BeginAddress).

docs/_docs/31-fasterkv-revivification.md
TedHartMS: Changed to:
The `VariableLenghtBlittableAllocator.GetValue` overload with (physicalAddress, physicalAddress + allocatedSize) that calls the default SpanByte implementation of `IVariableLengthStructureSettings` actually initializes the value space to be a valid SpanByte that includes the entire space (up to the maximum number of elements that can fit). However, other data types may not do this.

docs/_docs/31-fasterkv-revivification.md
TedHartMS: added:
- The largest bin is of size 16 * 2^16. This is because each FreeRecord holds only a long for storage, and 48 bits of that must be used for the address, leaving only 16 for the size. However, we scale the bin sizes by 16 since that is the minimum size.

docs/_docs/20-fasterkv-basics.md
TedHartMS: Updated in most recent edits: FASTER sets the tombstone.

docs/_docs/20-fasterkv-basics.md
TedHartMS: Updated in most recent edits:
1. That was a doc error; FASTER sets this and it is read-only to the app (setter is internal)
2. The app sets UpsertInfo.UsedValueLength or RMWInfo.UsedValueLength to the length required. FASTER will check to see if this is <= FullValueLength. If not, InPlaceUpdater returns false and FASTER ensures an appropriate allocation is made (or retrieved via revivification) before entering the `InitialUpdater` path.
- Item 2 applies to all similar situations: NeedInitialUpdate, NeedCopyUpdate, CopyUpdater, ConcurrentWriter

docs/_docs/20-fasterkv-basics.md
TedHartMS: Doc updated: this does go to InitialUpdater unless Canceled *also*; this combination means "delete it and return", so RMW returns (status.NotFound | status.Record.InPlaceUpdated).

docs/_docs/20-fasterkv-basics.md
TedHartMS: Doc Error; no record exists so no expiration is possible

docs/_docs/20-fasterkv-basics.md
TedHartMS: updated

docs/_docs/20-fasterkv-basics.md
TedHartMS: Current Delete() passes a ref to a default(Value) to hlog.GetRecordSize(). Changing this would require concomitant changes to record-length determination.

docs/_docs/20-fasterkv-basics.md
TedHartMS: doc error; fixed (see similar comment above)

docs/_docs/20-fasterkv-basics.md
TedHartMS: This comment applies to InPlaceUpdater, not CopyUpdater, which does not go to the IU path. 

Doc updated; if IPU sets both RMWInfo.CancelOperation and RMWInfo.DeleteRecord, then RMW sets the tombstone and returns (status.NotFound | status.Record.Created). Otherwise, .DeleteRecord moves to InitialUpdater.

The app may do its own internal "expire and reset" logic within the `InPlaceUpdater` call and return true, which to FASTER is indistinguishable from any other successful `InPlaceUpdater` call.

docs/_docs/20-fasterkv-basics.md
```diff
@@ -107,7 +107,104 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+    - `PostCopyUpdater`
+      - This is void because the record has already been CAS'd into the hash table, so it must perform operations that do not fail.
+
+**Upsert**
+
+  In all cases of **RequiredLength**, FASTER sets upsertInfo.FullValueLength to the full allocated length of the Value portion of the record.
+
+  - `ConcurrentWriter`
+    - ***Required Length***:
+      - If upsertInfo.UsedValueLength is sufficient, the app performs the update and returns true, and RMW returns (status.Found).
```
badrishc, line 184: should be upsertInfo.FullValueLength

badrishc:
Closing PR for now - we will target this for the next FASTER release (v2.1).

PR #654, opened 2022-02-18
[C#] Add complete pending to device API

badrishc:
So that we can ensure no outstanding IO when disposing allocator.

### DIFF
--- a/cs/src/core/Allocator/GenericAllocator.cs
+++ b/cs/src/core/Allocator/GenericAllocator.cs
@@ -191,6 +191,13 @@ namespace FASTER.core
         /// </summary>
         public override void Dispose()
         {
+            // Dispose base allocator
+            base.Dispose();
+
+            // Ensure pending IO on object log device is completed
+            objectLogDevice.CompletePending();
+
+            // Clean up data structures
             if (values != null)
             {
                 for (int i = 0; i < values.Length; i++)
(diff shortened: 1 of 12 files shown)


cs/src/core/Allocator/GenericAllocator.cs
```diff
@@ -191,6 +191,13 @@ internal override bool TryComplete()
         /// </summary>
         public override void Dispose()
         {
+            // Dispose base allocator
+            base.Dispose();
```
TedHartMS, line 195: It's generally better to destruct the base after derived. Did this need to move here?

TedHartMS (approved):

cs/src/core/Allocator/GenericAllocator.cs
```diff
@@ -191,6 +191,13 @@ internal override bool TryComplete()
...
+
+            // Ensure pending IO on object log device is completed
+            objectLogDevice.CompletePending();
```
tli2, line 198: This is probably a left over issue from before, but why does Dispose() here not call into dispose of objectLogDevice? It seems wrong to call CompletePending() without disposing, as that does not reliable stop future calls.

tli2 (approved):
Otherwise LGTM

cs/src/core/Allocator/GenericAllocator.cs
badrishc: We cannot dispose the device as we did not create the device, the user did. So they will have to dispose it.

badrishc:
Closing - instead we will update the device to catch and ignore exceptions in the uncommon race condition of IO after allocator dispose.

PR #668, opened 2022-03-02
[C#] Cancellation, ReadOptions, ReadFlags refactor

TedHartMS:


docs/_docs/20-fasterkv-basics.md
```diff
@@ -109,6 +109,74 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+  - ***Cancellation***: The app sets readInfo.CancelOperation True and returns false. Read() returns (status.Canceled).
+  - ***Other false return***: Read() returns (status.NotFound).
+
+**RMW** (`RMWInfo`)
+  - `InPlaceUpdater`
+    - ***Expiration***: The app sets rmwInfo.DeleteRecord true and returns false. FASTER will handle this as if the record was not found, and will enter the `NeedInitialUpdater` path, unless ***Cancellation*** was also done; in that case, FASTER will set recordInfo.Tombstone to true and will return (status.NotFound | status.InPlaceUpdated).
+      - The app may do its own internal "expire and reset" logic within the `InPlaceUpdater` call and return true, which to FASTER is indistinguishable from any other successful `InPlaceUpdater` call.
+    - ***Cancellation***: The app sets rmwInfo.CancelOperation True and returns false. If Expiration was also done, RMW() returns (status.NotFound | status.InPlaceUpdated); otherwise, RMW() returns (status.Canceled).
+    - ***Other false return***: Goes to `InitialUpdater` path.
```
badrishc, line 134: Do you mean `CopyUpdater` path here? If InPlaceUpdate returns false, we do a CopyUpdate.

docs/_docs/20-fasterkv-basics.md
```diff
@@ -109,6 +109,74 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+  - ***Cancellation***: The app sets readInfo.CancelOperation True and returns false. Read() returns (status.Canceled).
+  - ***Other false return***: Read() returns (status.NotFound).
+- `ConcurrentReader`
+  - ***Expiration***: The app sets readInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by in-place update to set the tombstone, and Read will return (status.NotFound | status.Record.InPlaceUpdated).
+  - ***Cancellation***: The app sets readInfo.CancelOperation True and returns false. Read() returns (status.Canceled).
+  - ***Other false return***: Read() returns (status.NotFound).
+
+**RMW** (`RMWInfo`)
+  - `InPlaceUpdater`
+    - ***Expiration***: The app sets rmwInfo.DeleteRecord true and returns false. FASTER will handle this as if the record was not found, and will enter the `NeedInitialUpdater` path, unless ***Cancellation*** was also done; in that case, FASTER will set recordInfo.Tombstone to true and will return (status.NotFound | status.InPlaceUpdated).
```
badrishc, line 131: If cancellation was not done, but DeleteRecord is set to true, IPU will still set the recordInfo.Tombstone bit before calling NIU, correct?

docs/_docs/20-fasterkv-basics.md
```diff
@@ -109,6 +109,74 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+      - ***Other false return***: Returns (status.NotFound) (but no status.Recored.Created).
+    - `PostInitialUpdater`
+      - This is void because the record has already been CAS'd into the hash table, so it must perform operations that do not fail.
+  - *CopyUpdater*
+    - `NeedCopyUpdate`
+      - ***Expiration***: The app sets rmwInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by setting the record's Tombstone and continuing with the CAS, and RMW will return (status.NotFound && status.Record.CreatedRecord).
+      - ***Cancellation***: The app sets rmwInfo.CancelOperation True and returns false. RMW() returns (status.Canceled).
+      - ***Other false return***: Returns (status.Found).
+    - `CopyUpdater`
+      - ***Expiration***: The app sets rmwInfo.DeleteRecord true and returns false. FASTER will perform a logical delete of the record by setting the record's Tombstone and continuing with the CAS, and RMW will return (status.NotFound && status.Record.CreatedRecord). This may be superseded by ***Cancellation***.
```
badrishc, line 152: On expiration, IPU deletes the record and goes through NIU path.
OTOH, CopyUpdater, on expiration, deletes the record, but does not go through NIU path.
Seems like an asymmetry here.

docs/_docs/20-fasterkv-basics.md
```diff
@@ -109,6 +109,74 @@ Apart from Key and Value, the IFunctions interface is defined on three additiona
...
+      - ***Cancellation***: The app sets rmwInfo.CancelOperation True and returns false. RMW() returns (status.Canceled). This supersedes ***Expiration***.
+      - ***Other false return***: Returns (status.Found) (but no status.Recored.Created).
+    - `PostCopyUpdater`
+      - This is void because the record has already been CAS'd into the hash table, so it must perform operations that do not fail.
+
```
badrishc, line 157: When an RMW operation encounters expiration (without cancellation), we need to retry_now the operation. While this most likely implies a NIU -> IU codepath, it does not necessarily imply so. If another thread managed to add a new value immediately after Tombstoning it, then our retry_now may end up doing a IPU or even a CU. This comment is really to confirm that this interpretation is correct.

docs/_docs/20-fasterkv-basics.md
badrishc: There are two options here:
* Basic correct version: CAS in a tombstone record, then retry_now the operation to perform the actual operation, most likely as NIU -> IU.
* Optimized version: Directly go to the NIU -> IU path and CAS in this value to the chain. If CAS fails, retry the operation. Potential semantic downside is that the chain does not explicitly record the expiration via a tombstone entry.

badrishc (approved):

PR #688, opened 2022-04-08
[C#] V2: Add StoreFunctions.Dispose

TedHartMS:
Move Dispose*() calls from separate IFunctions methods to a single Dispose(ref Key, ref Value, DisposeReason) on the new StoreFunctions type parameter to FasterKV


badrishc:
Closing as not part of v2.

PR #714, opened 2022-06-17
Fix snapshot checkpoint to scan only fuzzy region

badrishc:
Fix snapshot checkpoint to set startLogicalAddress to beginning of fuzzy region, instead of to FlushedUntilAddress
Fix snapshot checkpoint recovery to scan only fuzzy region

Also, make serializer take read locks with object log. While this is correct, an optimal solution would:
   - take no locks for non-fuzzy records (address < startLogicalAddress)
   - take no locks for fuzzy "v" version records (info.InNewVersion = false and address >= startLogicalAddress)
   - do not either lock or even serialize (i.e. serialize null) fuzzy "v+1" records (info.InNewVersion = true and address >= startLogicalAddress)

### DIFF
--- a/cs/src/core/Allocator/AllocatorBase.cs
+++ b/cs/src/core/Allocator/AllocatorBase.cs
@@ -1265,7 +1265,7 @@ namespace FASTER.core
         {
             if (Utility.MonotonicUpdate(ref SafeReadOnlyAddress, newSafeReadOnlyAddress, out long oldSafeReadOnlyAddress))
             {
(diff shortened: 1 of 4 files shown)


TedHartMS (approved):
Look good. This fixes the issue of CheckBucketVersionNew wrongly returning true for records with InVersionNew because _hybridLogCheckpoint.info.startLogicalAddress was set to faster.hlog.FlushedUntilAddress rather than GetTailAddress().

This does not change HLCI.headAddress and that's what SetRecoveryPageRanges returns for the headAddress of RecoveryOptions which is used for clearLocks. Similarly, this does not change the page parameters passed to ProcessReadSnapshotRange, and that means the clearLocks check will still override the start of range and process those pages.

PR #721, opened 2022-07-05
[C#] Add Stressapp

TedHartMS:
Add stress app and some related fixes:
- Initial work to clean up flushEvent handling for Async and add separate HandleOperationStatus overload for it. Addition work still to be done for RETRY_LATER issues discovered in stress.
- Add WithOutputs to InternalCompleteRetryRequests. 
- Add interfaces for epoch control and lock operations for (Lockable)UnsafeContext which don't need ephemeral types


cs/src/core/Utilities/Status.cs
```diff
@@ -135,6 +135,6 @@ public bool IsCompletedSuccessfully
         public byte Value => (byte)statusCode;
 
         /// <inheritdoc />
-        public override string ToString() => this.statusCode.ToString();
+        public override string ToString() => (this.Found ? "Found, " : string.Empty) + this.statusCode.ToString();
```
badrishc, line 138: why call out Found specifically?

badrishc (approved):

PR #753, opened 2022-10-06
[C#] FasterLog CommitAsync wait if no new commit done

badrishc:
If we await CommitAsync, and the operation does not start a new commit, we currently return immediately. Instead, we need to make sure commit has completed until the tail as of the CommitAsync call. Commit(spinWait: true) already ensures this, but the check was missing in CommitAsync.


tli2 (approved):
LGTM

PR #755, opened 2022-10-09
[C#] FasterLog scan fix when begin address moves ahead of scan

badrishc:
If scan is below begin address, update nextAddress before returning. Also, make sure currentAddress and outNextAddress are consistent w.r.t. shared nextAddress field.

### DIFF
--- a/cs/test/FasterLogScanTests.cs
+++ b/cs/test/FasterLogScanTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using System;
+using System.Threading;
 using FASTER.core;
 using NUnit.Framework;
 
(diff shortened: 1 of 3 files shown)


cs/test/FasterLogScanTests.cs
```diff
@@ -130,6 +131,57 @@ public void ScanBasicDefaultTest([Values] TestUtils.DeviceType deviceType)
             Assert.AreEqual(entryLength, currentEntry);
         }
 
+        [Test]
+        [Category("FasterLog")]
+        [Category("Smoke")]
+        public void ScanBehindBeginAddressTest([Values] TestUtils.DeviceType deviceType)
+        {
+            // Create log and device here (not in setup) because using DeviceType Enum which can't be used in Setup
+            string filename = path + "LogScanDefault" + deviceType.ToString() + ".log";
```
TedHartMS, line 140: You actually can use that parameter in Setup():

            DeviceType deviceType = DeviceType.LocalMemory;
            foreach (var arg in TestContext.CurrentContext.Test.Arguments)
            {
                if (arg is TestUtils.DeviceType argDeviceType)
                    deviceType = argDeviceType;
            }
            device = CreateTestDevice(deviceType, ...);

TedHartMS (approved):

PR #768, opened 2022-11-15
Fix: Some platforms try get a segmentId from the main log file.

### LINKED ISSUE #767: RecoverFiles fails to int.Parse on bareName nuget v.2.1.0 using .net7 on windows 10
Thanks for the amazing Library! It's a very impressive work.

I'm trying to recover files after shutting down.

In my log directory I have:
```
logfile
logfile.0
```
The logic for populating the segids is picking up `logfile` and trying to parse an end off of the back of it. It seems like it the expectation is that di.GetFiles("*") would only return files that a have `logfile` as the beginning portion, but should exclude `logfile`.

I end up with an int.Parse error as it tries to parse an empty string from:
```c#
            foreach (System.IO.FileInfo item in di.GetFiles(bareName + "*"))
            {
                int segmentId = Int32.Parse(item.Name.Replace(bareName, "").Replace(".", ""));
                ...
            }
```
I think we should add an extra check inside the loop to continue if we end up with an empty segment number string.

eg:
```c#
// (Managed)LocalStorageDevice.cs :: private void RecoverFiles()
            foreach (System.IO.FileInfo item in di.GetFiles(bareName + "*"))
            {
                // on .net7 windows 10 we are getting bareName as one of the files returned from DirectoryInfo.GetFiles
                // the file that is bareName has no ".{segmentId}" appended, so the int32.parse below would fail.
                if (item.Name == bareName)
                {
                    continue;
                }
                
                int segmentId = Int32.Parse(item.Name.Replace(bareName, "").Replace(".", ""));
                
                ...
            }
```

Would you like me to submit a PR?

Astn:
Fixes #767 

Avoids `Int32.Parse` failure on platforms that return the main log file as one of the matches for `di.GetFiles(bareName + "*")`.

### DIFF
--- a/cs/src/core/Device/LocalStorageDevice.cs
+++ b/cs/src/core/Device/LocalStorageDevice.cs
@@ -153,6 +153,10 @@ namespace FASTER.core
             List<int> segids = new();
             foreach (System.IO.FileInfo item in di.GetFiles(bareName + "*"))
             {
+                if (item.Name == bareName)
+                {
+                    continue;
+                }
                 segids.Add(int.Parse(item.Name.Replace(bareName, "").Replace(".", "")));
             }
             segids.Sort();
--- a/cs/src/core/Device/ManagedLocalStorageDevice.cs
+++ b/cs/src/core/Device/ManagedLocalStorageDevice.cs
@@ -72,6 +72,10 @@ namespace FASTER.core
             List<int> segids = new();
             foreach (FileInfo item in di.GetFiles(bareName + "*"))
             {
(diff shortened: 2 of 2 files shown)


Astn:
@microsoft-github-policy-service agree

badrishc:
Would suggest we stick to the if/continue version, instead of using Linq, which is generally more expensive so we have avoided it in the source code. Thanks.

Astn:
Sounds good. I had the same thought, but wasn't sure which way you world prefer it. I've reverted it back to the if/continue version.

PR #805, opened 2023-03-16
Adding Research Code from DPR & DARQ

tli2:
Prototype implementation for DPR and DARQ work.


cs/remote/src/FASTER.common/MessageType.cs
```diff
@@ -71,5 +71,25 @@ public enum MessageType : byte
         /// Pending result
         /// </summary>
         PendingResult,
+        
+        /// <summary>
+        /// DARQ Enqueue
+        /// </summary>
+        DarqEnqueue,
```
badrishc, line 78: DARQ specifics leaking into core/common

cs/remote/src/FASTER.common/SimpleObjectPool.cs
```diff
@@ -12,20 +11,26 @@ namespace FASTER.common
     /// Object pool
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    public class SimpleObjectPool<T> : IDisposable where T : class, IDisposable
+    public class SimpleObjectPool<T> : IDisposable where T : class
     {
         private readonly Func<T> factory;
+        private readonly Action<T> destructor;
```
badrishc, line 17: why this change? there is other code we have that depends on this.

cs/remote/src/FASTER.common/WireFormat.cs
```diff
@@ -23,6 +23,10 @@ public enum WireFormat : byte
         /// Similar to DefaultVarLenKV but with WebSocket headers (binary)
         /// </summary>
         WebSocket = 2,
+        
+        DarqSubscribe,
```
badrishc, line 27: are these really new wire format level items? and DARQ doesnt seem like somthing to be exposed in this level.

cs/remote/src/FASTER.common/NetworkSenderBase.cs
```diff
@@ -56,10 +58,7 @@ public NetworkSenderBase(int serverBufferSize)
         public abstract bool SendResponse(int offset, int size);
 
         /// <inheritdoc />
-        public abstract void SendResponse(byte[] buffer, int offset, int count, object context);
```
badrishc, line 59: i believe these abstracts are needed as they are used by other codes not in this repo.

cs/remote/src/FASTER.common/SimpleObjectPool.cs
tli2: Needed to accommodate objects (e.g., SEAAs) that needs custom Dispose() logic. I have changed the default argument ordering so it should now be a backward compatible API change for existing code.

cs/remote/src/FASTER.common/NetworkSenderBase.cs
tli2: Reverted

cs/remote/src/FASTER.common/WireFormat.cs
tli2: Same as above for MessageType

PR #821, opened 2023-04-19
[C#] State machine adjustments

badrishc:
* Make sure ongoing multi-key ops drain out in PREPARE
* Correctly handle index resize


tli2 (commented):
LGTM

PR #827, opened 2023-04-30
Use Microsoft.Extensions.Logging.Abstractions instead of Microsoft.Extensions.Logging to reduce dependencies.

Summpot:
<del>Fixes #717 because Microsoft.Extensions.Logging.Abstractions has very few dependencies.</del>
#717 appears to have been fixed elsewhere, and now references to higher versions of Microsoft.Extensions.Logging.Abstractions work properly in the Microsoft.Azure.Functions.Worker.Sdk project.


badrishc:
We need the .nuspec for our CI auto-release based on $version in there, so it should not be removed.

badrishc:
We can look into whether using this instead of $version is possible in our internal release pipeline. But generally would rather not fix something that isn't broken.

PR #892, opened 2023-12-04
[C#] Clarify and enforce BumpEpoch protection assumptions

### LINKED ISSUE #872: Log tail iteration does not end when log completed
The documentation seems to suggest (in the [Log Completion](https://microsoft.github.io/FASTER/docs/fasterlog-basics/#log-completion) section), that an iterator can check for log completion with the `Ended` property.

However when a log is completed, it seems that the iterator's `WaitAsync` method does not return during tail iteration in order for this property to be checked.

In this example code, the 'Scan complete' message is never sent to the console, and the scanning iterator waits forever after the log is completed.

```C#
using System.Diagnostics;
using FASTER.core;

using var settings = new FasterLogSettings("./Test", deleteDirOnDispose: true)
{
    AutoRefreshSafeTailAddress = true
};
using var log = new FasterLog(settings);

byte[] buffer = new byte[2048];
Random.Shared.NextBytes(buffer);

await Task.WhenAll(EnqueueThread(), ScanThread());
return;

async Task EnqueueThread()
{
    for (int count = 0; count < 5; ++count)
    {
        await log.EnqueueAsync(buffer);
        await Task.Delay(1000);
    }

    log.CompleteLog();
    Console.WriteLine("Enqueue complete");
}

async Task ScanThread()
{
    using var iterator = log.Scan(log.BeginAddress, long.MaxValue, scanUncommitted: true);
    while (true)
    {
        byte[] result;
        while (!iterator.GetNext(out result, out _, out _))
        {
            if (iterator.Ended)
            {
                Console.WriteLine("Scan complete");
                return;
            }

            await iterator.WaitAsync();
        }
    
        Console.WriteLine("Received buffer");
        Debug.Assert(result.SequenceEqual(buffer));
    }
}
```

Is the correct usage in this case that the iterator's `WaitAsync` method requires a `CancellationToken` passed in that is cancelled separately at the time the log is completed? The documentation and sample don't suggest this.

Another minor issue: the documentation also mentions the `void RefreshUncommitted(bool spinWait = false)` method, which is no longer part of the API, but instead seems to be configured in the `FasterLogSettings` (as in the code sample above).

tli2:
Revise BumpEpoch to accommodate calling from both protected and unprotected contexts. Revise FasterLog completion logic to avoid epoch double-protect. From now on, users should assume that any trigger action associated with BumpEpoch will execute on a protected thread after it is safe. Caller thread of BumpEpoch can be either protected or unprotected.

fix #872

### DIFF
--- a/cs/src/core/Epochs/LightEpoch.cs
+++ b/cs/src/core/Epochs/LightEpoch.cs
@@ -181,7 +181,7 @@ namespace FASTER.core
         }
 
         /// <summary>
-        /// Enter the thread into the protected code region
+        /// Enter the thread into the protected code region.
         /// </summary>
         /// <returns>Current epoch</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -225,7 +225,6 @@ namespace FASTER.core
         /// <returns></returns>
         long BumpCurrentEpoch()
         {
-            Debug.Assert(this.ThisInstanceProtected(), "BumpCurrentEpoch must be called on a protected thread");
             long nextEpoch = Interlocked.Increment(ref CurrentEpoch);
 
             if (drainCount > 0)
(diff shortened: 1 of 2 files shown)


TedHartMS (approved):
will make a couple changes after merge to feature branch

PR #893, opened 2024-01-12
patch bug in clearing of index bits during recovery

sebastianburckhardt:
This bug was already fixed in the latest release, but Netherite is still on an older release. 
We confirmed that this fix does resolve a hang we were able to repro after receiving details from this customer: 
https://github.com/Azure/azure-functions-durable-extension/issues/2603

The original fix is here: https://github.com/microsoft/FASTER/commit/a6c0dfa5f96dc0349987b3035e73fe90faddbfa4


badrishc:
Closing PR in favor of hotfix for the older version of FASTER via branch.

PR #901, opened 2024-03-23
use IsReferenceOrContainsReferences for .NET Standard 2.1+ and 6_0+

Tornhoof:
This ports the change done by @paulusparssinen done in https://github.com/microsoft/garnet/pull/36 to FASTER.
This only applies to the builds for .NET STANDARD 2.1 and 6_0+ as otherwise the helper method is not available, the reflection method is also still used in the code.

### DIFF
--- a/cs/src/core/Utilities/Utility.cs
+++ b/cs/src/core/Utilities/Utility.cs
@@ -154,6 +154,10 @@ namespace FASTER.core
         /// <returns></returns>
         internal static bool IsBlittable<T>()
         {
+#if NETSTANDARD2_1_OR_GREATER || NET6_0_OR_GREATER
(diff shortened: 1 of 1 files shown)


badrishc (approved):

PaulusParssinen:
Came across this port and I believe relying on dummy `GCHandle` to check blittability is invalid on .NET 7+ after https://github.com/dotnet/runtime/pull/68694 (see also original issue https://github.com/dotnet/runtime/issues/68686)

Am I correct @jkotas ?

PaulusParssinen:
Deleted earlier incorrect comment, misread the changes 🤦‍♂️ 

This actually fixes a bug as the .NET 7+ onwards it is invalid to rely on dummy `GCHandle` for the blittability check (see https://github.com/dotnet/runtime/pull/68694)

PR #909, opened 2024-03-31
#908 Optimize dependencies

### LINKED ISSUE #908: Remove polyfil dependencies for latest TFM
**Is your feature request related to a problem? Please describe.**
I want to minimise dependencies in my project by utilising framework dependencies wherever possible

**Describe the solution you'd like**
The System.Text.Json package becomes conditional so only added for the older frameworks

**Describe alternatives you've considered**
Accept the additional dependency

**Additional context**
n/a

thompson-tomo:
closes #908

### DIFF
--- a/cs/src/core/FASTER.core.csproj
+++ b/cs/src/core/FASTER.core.csproj
@@ -51,10 +51,6 @@
     <PackageReference Include="System.Interactive.Async" Version="6.0.1" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' != 'netstandard2.1'">
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
-  </ItemGroup>
-
   <Choose>
     <When Condition="'$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'netstandard2.1'">
       <ItemGroup>
--- a/cs/src/core/FASTER.core.csproj
+++ b/cs/src/core/FASTER.core.csproj
@@ -20,7 +20,7 @@
     <StartupObject />
     <AssemblyOriginatorKeyFile>../../FASTER.snk</AssemblyOriginatorKeyFile>
     <DelaySign>false</DelaySign>
-	<EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <DocumentationFile>bin\$(Platform)\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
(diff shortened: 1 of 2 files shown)


cs/src/core/FASTER.core.csproj
```diff
@@ -29,45 +29,48 @@
...
-  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.1'">
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="System.Memory" Version="4.5.5" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
     <PackageReference Include="System.Interactive.Async" Version="6.0.1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' != 'netstandard2.1'">
```
TedHartMS, line 54: I think you meant "== netstandard2.1"

thompson-tomo: It looks that dependency is not even needed hence why it was building fine, so have removed it,

cs/src/core/FASTER.core.csproj
```diff
@@ -29,45 +29,44 @@
...
   </ItemGroup>
-  
-  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.1'">
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="System.Memory" Version="4.5.5" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
     <PackageReference Include="System.Interactive.Async" Version="6.0.1" />
   </ItemGroup>
```
TedHartMS, line 52: This is failing a local build due to IAsyncEnumerable not implementing FirstAsync, so I think System.Interactive.Async is required

badrishc:
Could this pull request apply to Garnet as well (Tsavorite)?

TedHartMS:
> Could this pull request apply to Garnet as well (Tsavorite)?

No; we don't include these there (and don't switch on TargetFramework)

cs/src/core/FASTER.core.csproj
thompson-tomo: Ahh, pinpointed the issue which is actually in AzureStorageDevice. In the core it was set to not be included for NetStandard2.1 and then in Azure it got explicitly added. As such i have cleaned this up. And all projects now built.

thompson-tomo:
Still changes are missing.

PR #952, opened 2025-11-25
Add dotnet 10 build target

Arithmomaniac:
The dotnet 10 build target removes the `System.Linq.Async` import, preventing the sort of problems from consumers mentioned by https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/asyncenumerable

### DIFF
--- a/azure-pipelines-full.yml
+++ b/azure-pipelines-full.yml
@@ -39,6 +39,12 @@ jobs:
       packageType: 'sdk'
       version: '7.0.x'
 
+  - task: UseDotNet@2
+    displayName: Use .NET 10.0
+    inputs:
+      packageType: 'sdk'
+      version: '10.0.x'
+
   - task: NodeTool@0
     inputs:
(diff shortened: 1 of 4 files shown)


Arithmomaniac:
System.Linq.Async added a new backwards-compatible version that users can install as a non-transitive dependency to avoid these issues:

> In most cases the fix is simple: upgrade to System.Linq.Async v7. (If you were using it directly, this just means upgrading your existing package reference to the latest version. If you have ended up with an indirect reference through some other package you'll need to add a new reference to the latest version of System.Linq.Async.) In some cases, for reasons described later, you might need to add a reference (or upgrade an existing reference) to System.Interactive.Async v7.
>
> https://endjin.com/blog/2025/11/ix-v7-dotnet-10-linq-iasyncenumerable

It would be nice to take care of this for users automatically with a PR like this one but it's not pressing enough to automatically warrant a new build target.

